PR #1, opened 2024-01-07
Add steam support for the loader

DeonPoncini:
After decompiling the steam game (see
https://github.com/DeonPoncini/swep1r_decomp) I was able to compare all the functions that are currently being patched in for the loader with what is being done within the GOG version

They appear to be the same:
https://docs.google.com/spreadsheets/d/1KqYdug7rk9F_bZ3RHs9HiPzea73Ovsd5Jkctcmt4zbM/edit?usp=sharing is an analysis of these showing the virtual addresses in each version, which aligns

I was able to build the rewrite DLL and run the steam game successfully

### DIFF
--- a/loader/loader.cpp
+++ b/loader/loader.cpp
@@ -7,26 +7,31 @@
 
 std::string GetLastErrorAsString()
 {
-    //Get the error message ID, if any.
     DWORD errorMessageID = ::GetLastError();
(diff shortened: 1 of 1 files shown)


tim-tim707:
Removed some comments and refactored md5 matching. Thanks for the PR !

PR #2, opened 2024-01-07
Compile fixes

DeonPoncini:
I have updated the scripts to make the base code compile, using a default g++ compiler on Windows 11

Main fixes:

1) there were a lot of missing forward declarations, rearranged things to remove issues with no definition, or partial types being used not as pointers before declaration
2) -fpermissive was required on the compiler
3) roll/pitch/yaw on on tr_rot is all one variable (an rdVector3) convert those
4) UUID vs GUID issues
5) incorrectly generated globals
6) some redefined types that are included by existing headers

### DIFF
--- a/src/Primitives/rdMatrix.c
+++ b/src/Primitives/rdMatrix.c
@@ -239,7 +239,7 @@ void rdMatrix_ExtractTransform(rdMatrix44* mat, swrTranslationRotation* tr_rot)
         {
             fVar4 = 0.0;
         }
-        tr_rot->yaw = fVar4;
+        tr_rot->yaw_roll_pitch.x = fVar4;
     }
     else
     {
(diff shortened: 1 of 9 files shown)


src/Primitives/rdMatrix.c
```diff
@@ -854,7 +855,7 @@ void rdMatrix_PostMultiply34(rdMatrix34* mat1, rdMatrix34* mat2)
 void rdMatrix_PreRotate34(rdMatrix34* out, rdVector3* rot)
 {
     rdMatrix34 tmp;
-    rdMatrix_BuildRotate34(out, &tmp);
```
tim-tim707, line 857: Why comment this ? I believe there is a call to BuildRotate

DeonPoncini: tmp is of type rdMatrix34, but rdMatrix_BuildRotate34's second parameter expects a rdVector3

As I was just trying to get it compiling and this code wasn't loaded, I didn't try to see how I could convert an rdMatrix34 to an rdVector3 (well other than doing a matrix multiply to select a row vector)

tim-tim707: Indeed a mistake on my end. Thanks !

PR #3, opened 2024-03-10
Experiments

tly000:
big changes: 

- added CMakeLists.txt to build most of "src" as a static "swe1r_functions" library and build dinput_hook
- dinput_hook that hooks all functions from swe1r_functions and optionally replaces them
     - contains imgui with directx6 backend
     - uses detours library for hooking
     - adds some hooks to fix duplicated mouse cursor in windowed mode
- globals.h now point to the data of the executable
     - globals.c is unused
- compile fixes and address fixes in various files
- model reversing in swrModel.h/.c and types_model.h

if you dont want the dinput_hook, cmake stuff or anything else, let me know 🙂


tim-tim707:
Impressive work indeed !

> if you dont want the dinput_hook, cmake stuff or anything else, let me know 🙂
I'd like the project to have minimal dependencies overall, the best being a single C/C++ compiler and a compile script (.bat, .sh for each platform).

About Cmake and Makefile: Is it necessary ? Can we do without it ? I know that its powerful and works well with IDEs but if possible I'd like it not to be in the project. Even if it's unlikely that any user will try to compile the code at home, it would be better if it were as simple as possible, ie Running `compile.bat|sh` and be done with it.

I think detour is a big piece of the puzzle we can't do without ? Especially for loops at the beginning of functions I think. OpenJKDF2 doesn't use Detour and they managed to go far without it. I'm mostly OK to have it as a dependency.

Since its being compiled as a static library, we can't use it with a loader right ? I still want this option to be available for when the time will be to compile the `swr_reimpl.dll` to just be a drag and drop in the mod folder, and then run the mod loader.

About the dinput_hook folder and all the new functionalities: Is there a way to add a compile flag and macros, or another process in order to separate dev code and tooling from the reimplementation code ? (Or I miss understand how the hooking works / is suppose to work in the end)

I also noticed the hardcoded path of dlls here in this cmake: `dinput_hook/CMakeLists.txt`. I think we should add these dlls to our project instead, it would be more self-contained. Like a version pinning but without package manager. It would help with reproducibility.

All the reversed stuff, compile fixing and update to the global generation script are awesome and I could probably merge them immediately. It doesn't seem like I can accept parts of a PR to merge directly so we'll do it once everything is ok

tly000:
> About Cmake and Makefile: Is it necessary ? Can we do without it ? I know that its powerful and works well with IDEs but if possible I'd like it not to be in the project. Even if it's unlikely that any user will try to compile the code at home, it would be better if it were as simple as possible, ie Running `compile.bat|sh` and be done with it.
> 

yep, i think i could remove it, or make it optional. do you mind if i convert my CMakeLists.txt to a makefile? everyone who has the capacity to build this (i.e. has a mingw installed) also has make installed automatically in the mingw distribution. 

> I think detour is a big piece of the puzzle we can't do without ? Especially for loops at the beginning of functions I think. OpenJKDF2 doesn't use Detour and they managed to go far without it. I'm mostly OK to have it as a dependency.
> 

i think the main advantage of detours is, that you can still call the original function, because its backed up. this makes it easier to inject stuff. otherwise it should not make a big difference as far as i know. 

> Since its being compiled as a static library, we can't use it with a loader right ? I still want this option to be available for when the time will be to compile the `swr_reimpl.dll` to just be a drag and drop in the mod folder, and then run the mod loader.
> 

if you mean the "swe1r_functions" library, this is mostly a convenience feature, i.e. building sweir_functions with different compile flags than the rest. this does not change much in the actual dll.

> About the dinput_hook folder and all the new functionalities: Is there a way to add a compile flag and macros, or another process in order to separate dev code and tooling from the reimplementation code ? (Or I miss understand how the hooking works / is suppose to work in the end)
> 

yep, i think this whole hooking process is very experimental, it needs debug symbols and the bfd library to parse the debug symbols in the dll file. for a "production build" (😛) it would be best to change this, but i dont have a plan for that right now. do you have any recommendations how to do it? maybe we can just use your "hook.c" file for this case? 

> I also noticed the hardcoded path of dlls here in this cmake: `dinput_hook/CMakeLists.txt`. I think we should add these dlls to our project instead, it would be more self-contained. Like a version pinning but without package manager. It would help with reproducibility.
> 

yes, basically all those files are available in every mingw version and i dont really want to add them to git because i dont know how stable their abi is. this means that it might only work with exactly my mingw build (https://winlibs.com, latest x86 posix threads build with UCRT). instead, the hardcoded paths must be replaced by code that tries to find the libraries/dlls.

tim-tim707:
> everyone who has the capacity to build this (i.e. has a mingw installed) also has make installed automatically in the mingw distribution.

Cool then, let's do it !

> i think the main advantage of detours is, that you can still call the original function, because its backed up. this makes it easier to inject stuff. otherwise it should not make a big difference as far as i know.

Yeah its a great feature, especially if we can have hooking dynamically on top, it would be much better for developpement.

> if you mean the "swe1r_functions" library, this is mostly a convenience feature, i.e. building sweir_functions with different compile flags than the rest. this does not change much in the actual dll.

Ok for me then 👍 

>  maybe we can just use your "hook.c" file for this case?

That is the plan for release yes. We probably don't need interactive hooking in this case and just let the modloader do its magic.

> (https://winlibs.com/, latest x86 posix threads build with UCRT). instead, the hardcoded paths must be replaced by code that tries to find the libraries/dlls.

If you can choose a specific version (version 5 I suppose) and add it to the readme instructions with the link it should be good enough for "version pinning". We probably won't need to change version ever and I suppose the available version will still be available in the future.

Is there anything you want to add before I merge ?

tly000:
> Is there anything you want to add before I merge ?

should i fix the things that you mentioned before the merge or some other time?

tim-tim707:
I would prefer if adding a notice in the readme and removing CMake was done before merging.

tly000:
it should be ready now to merge 👍

PR #4, opened 2024-03-12
Add a script to calculate % completion

DeonPoncini:
This script calculates percentage completion of identification of functions

This works by using the existing GenerateMasterHeader.py script to create a list of all functions already known

This introduces a new Ghidra script, GenerateFunctionList.py, that creates a list of functions from the decompilation.

Then these lists are compared to see how many are known and provide a completion percentage


tim-tim707:
Looks good to me !

PR #7, opened 2024-03-16
Automatically generate globals.h/.c on build

DeonPoncini:
Automatically execute the python script to generate globals if data_symbols.syms changes, so we get a new global header generated


scripts/GenerateGlobalHeaderFromSymbols.py
```diff
@@ -55,18 +59,19 @@
...
 
-with open("./src/globals.h", "w", encoding="ascii") as file:
+with open(output_file_h, "w", encoding="ascii") as file:
     file.write(rendered_output)
 
-template_file = "./src/globals.c.j2"
-env = Environment(loader=FileSystemLoader("."))
-template = env.get_template(template_file)
+env = Environment(loader=FileSystemLoader(project_root))
+template = env.get_template("src/globals.c.j2")
```
tim-tim707: "src" + os.sep + "globals.c.j2" if we want to be consistent. Or use os.path.join

CMakeLists.txt
```diff
@@ -5,11 +5,22 @@ set(CMAKE_CXX_STANDARD 20)
...
 
 add_compile_options(
-    -Werror=incompatible-pointer-types
-    -Werror=implicit-function-declaration
+    #-Werror=incompatible-pointer-types
+    #-Werror=implicit-function-declaration
+    #-w
+    -g
     -mfpmath=387
 )
```
tim-tim707: Can we still use warning flags please. like -Wall and -Werror. Even if its not -Werror so that it doesn't block compilation, it would be better long term, such as explicit casting and signed / unsigned comparisons

tly000: those two flags find a lot of typical c problems, this is why i used them. i fixed ~20 bugs using those flags.

tly000: and about '-g': the correct way to enable debug symbols in cmake is `-DCMAKE_BUILD_TYPE=Debug` via command line or `set(CMAKE_BUILD_TYPE Debug)` at the beginning of the CMakeLists.txt

DeonPoncini: I actually did not mean to submit this part of the change, I will revert it.

scripts/GenerateGlobalHeaderFromSymbols.py
DeonPoncini: Unfortunately https://github.com/pallets/jinja/issues/767 see this issue, I have cleaned it up so if we want to implement a workaround we can

PR #8, opened 2024-03-16
Remove globals.c/globals.h as autogenerated code

DeonPoncini:
These are automatically generated from data_symbols.syms, so do not keep them in the repository. They will be autobuilt from the cmake build


tim-tim707:
Will merge with previous one. Can you add a little comment in the templates to precise its generated at compile time ? Thanks !

DeonPoncini:
Yes, I will update the README as well

PR #10, opened 2024-03-16
Fix rdMatrix multiply44 implementation to work

DeonPoncini:
rdMatrix_Multiply44 and rdMatrix_Multiply44Acc were not functioning correctly because the input pointers could be aliases of each other - and they were in some cases identified (e.g. FUN_004819b0 calling rdMatrix_Multiply44 aliases out and mat2)

Copy matrices locally first before including in the update

Also, preserve the exact multiplication order from the disassembly


tim-tim707:
Looks about right 👍

PR #11, opened 2024-03-17
Autogenerate bidirectional hooks on compile

DeonPoncini:
This introduces a script that automatically generates all hooks by reading through the source code, and looking for methods with a comment like:

// 0x<address> HOOK
<function declaration>

It must be in this exact format, with the function on a new line. This will then generate an entry into hook_generated.c that hooks the address mentioned to that function

For other functions that just have:

// 0x<address> (anything else)
<function declaration>

A reverse hook will be created, hooking from the function itself back into the executable, so those functions can be used before they are implemented, allowing for partial decompilation (just a function and not the functions it calls)

This is an attempt to bring together the work done for the basic hooking in the launcher, and the dynamic hooking within dinput_hook.

Advantages:
No external depedendencies (no need for detours or BFD) Generates a file that allows for auditing, so its clear what is getting hooked
Works in both debug and release builds
Runs automatically every time the source files change Works for the launcher as well as dinput method for dependency Maintains a log for hooks

Disadvantages:
Requires running the cmake build after each change to hook, BFD allowed this to happen magically/automatically without running the build Does not allow calling the original function when hooked, although likely this can be implemented


scripts/GenerateHooks.py
tim-tim707, line 1: Please use a jinja template instead of writing the file line by line with `file.write()`. It shouldn't be too difficult since you already have all the infos you need to fill the file. There are for-loops in jinja as well if needed.

DeonPoncini: time to learn how to use jinja. I was thinking after this gets put in, to move all the generated files to a src/generated or top level /generated that can have a .gitignore that just ignores everything, we can also move the .j2 files to a separate src/template or /template folder. This is for a later clean up patch however

src/dllMainExp.c
tim-tim707, line 1: What is the difference between this file and the dllMain.c above ?

DeonPoncini: I should rename this - I just called it exp for experimental but really should be dllMainDinput or something (open to suggestions)

This replaces dinput.dll, whereas the other dllMain.c creates just swp_reimpl.dll patched in with the loader. I created this so you could hook in through dinput.dll. Both methods work however with this solution and you can use either dll

Right now, swr_reimpl is only buildable from the compile.bat, and reimplementing all the cmake magic there is difficult. I could instead build loader in cmake if you prefer, but either way we want both methods available

scripts/GenerateHooks.py
DeonPoncini: I am glad I learned jinja, made both my python cleaner and the output file, updated and fixed now

src/hook_generated.h
```diff
@@ -0,0 +1,15 @@
+#ifndef HOOK_GENERATED_H
+#define HOOK_GENERATED_H
+#include <stdio.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+void hook_generated(FILE* hook_log);
```
DeonPoncini: Do you think we should put this directly into hook.h? Probably cleaner

src/dllMainExp.c
DeonPoncini: Renamed in the new patch

src/hook_generated.h
tim-tim707: Agree 👍

DeonPoncini: I have pushed a new patch with this change

PR #13, opened 2024-03-19
Fix pointer aliasing issues within rdMatrix.c

DeonPoncini:
The rdMatrix family of functions were not safe for being used with the same input and output parameters, which is done in a number of places, causing them to provide the wrong outputs. Make the functions safe for aliased pointers

Also, the rdMatrix_Copy44 implementation was incorrect, replaced it with a simple memcpy with the same behavior


tim-tim707:
Does the game originally have all these `memcpy`s ? If not, can you add a comment `// Added` before lines not in the original game ? Thanks !

DeonPoncini:
I have surrounded all changes with DELTA: <explanation> END DELTA

This is to make it greppable, as well as in the future we could add like #define BIT_EXACT or something to remove these parts and replace it with something bit exact if we end up going down that route

tim-tim707:
Thats a better way of doing things 👍

PR #15, opened 2024-03-23
Fix multiplayer compile errors under cmake

DeonPoncini:
Most files when referencing other headers use #include <Folder/header.h>. This was changed in the latest update for this file to include raw header includes

Use this style over including all headers in the cmake build, unless the intention is to change to this style everywhere

### DIFF
--- a/src/Swr/swrMultiplayer.c
+++ b/src/Swr/swrMultiplayer.c
@@ -3,8 +3,8 @@
 #include "globals.h"
 #include "macros.h"
 
-#include <stdComm.h>
-#include <sithMulti.h>
+#include <Win95/stdComm.h>
+#include <Dss/sithMulti.h>
 
 // 0x00412640
 void swrMultiplayer_SetInMultiplayer(int bInMultiplayer)


src/Swr/swrMultiplayer.c
```diff
@@ -3,8 +3,8 @@
 #include "globals.h"
 #include "macros.h"
 
-#include <stdComm.h>
-#include <sithMulti.h>
+#include <Win95/stdComm.h>
+#include <Dss/sithMulti.h>
```
DeonPoncini, line 8: @tim-tim707 is the intention to go to this lack of path style on includes going forward? I see the changes you made to compile.bat, but most files still use <Win95/header.h> style instead

I have an alternative fix redoing the cmakelists, but went for this first

tim-tim707:
Haven't thought about it that much but specifying path is probably better in the long run

DeonPoncini:
I agree with you that is my recommendation

DeonPoncini:
Is this ok to merge?

PR #16, opened 2024-03-26
Decompile swrText_CreateEntry

DeonPoncini:
Provide an implementation of swrText_CreateEntry (0x004503e0) and identify child function as a custom sprintf implementation, dubbed as "swrText_Sprintf"

Implement format strings as constants as they are present in the system memory like this being reused

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -345,6 +345,9 @@ swrTextEntries1Count 0x0050c750 int # < 0x80
 
 swrTextEntries2Count 0x0050c758 int # < 0x20
 
+swrTextFmtString1 0x004b2304 char[4] = "%s\0\0"
+swrTextFmtString2 0x004c3e48 char[8] = "~f%d%s\0\0"
+
 rdMatrixStack34_size 0x0050c6f4 int
 
 debug_showSurfaceFlags 0x0050c88c int
--- a/src/Swr/swrText.c
+++ b/src/Swr/swrText.c
@@ -84,13 +84,62 @@ char* swrText_Translate(char* text)
     return NULL;
 }
 
-// 0x004503e0
-void swrText_CreateEntry(short x, short y, char r, char g, char b, char a, char* screenText, int formatInt, int isEntry2)
-{
-    HANG("TODO, easy");
+// 0x0049eb80
+int swrText_Sprintf(char *str, const char *format, ...) {
+    HANG("TODO");
+    return 0;
 }
 
-// 0x00450530
+// 0x004503e0 HOOK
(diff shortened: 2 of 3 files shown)


data_symbols.syms
```diff
@@ -345,6 +345,9 @@ swrTextEntries1Count 0x0050c750 int # < 0x80
 
 swrTextEntries2Count 0x0050c758 int # < 0x20
 
+swrTextFmtString1 0x004b2304 char[4] = "%s\0\0"
+swrTextFmtString2 0x004c3e48 char[8] = "~f%d%s\0\0"
+
 rdMatrixStack34_size 0x0050c6f4 int
```
DeonPoncini, line 362: @tim-tim707 this feels somewhat unnatural. I did this as in the decomp, it was pointing at DAT_ values with these strings present

However, in natural coding I would likely just type in the string as a constant. I tried to follow the decomp, but wanted to ask a style question when encountering these format strings (there are a number present)

tim-tim707: I dont really know what would be best here. I personnaly like righting the string constant like in C but it wont be in the right location in the executable afterward. I dont think its important since the data is the same and the location irrelevant for function. It will still make matching decomp harder later on. Maybe we can do for now and worry about matching later.

DeonPoncini: Ok, so I think if that is the case, lets go with what I have here and I will follow this convention - we can clean up these globals when we are free from hooking in the future

src/Swr/swrText.c
```diff
@@ -84,13 +84,62 @@ char* swrText_Translate(char* text)
     return NULL;
 }
 
-// 0x004503e0
+// 0x0049eb80
+int swrText_Sprintf(char *str, const char *format, ...) {
```
DeonPoncini: This should just be sprintf as its the stdlibrary function defined in swr_stdlib.h - will update when the other patch is up allowing for calling standard library functions

tim-tim707: It's already defined in `swr_stdlib.h` yes, like most of the static stdlib. I don't get what you want to do exactly ?

DeonPoncini: Yeah sorry I made this comment after I did the commit, as I realized exactly what you said - that I should have just called sprintf. I didn't look at swr_stdlib.h until after I figured out this was sprintf

I will put up another patch just using sprintf. The other patch I mentioned was this one https://github.com/tim-tim707/SW_RACER_RE/pull/18 but sprintf is one stdlib function that doesn't crash if you use the msvcrt ones (unlike the file ones which do crash)

DeonPoncini: I have updated this to just use sprintf instead. This should be good to merge now

PR #17, opened 2024-03-26
Test every function independently, enable and annotate

DeonPoncini:
To generate this patch, every function was taken independently, enabled one a time to see which caused issues with the game. The results are found here:

https://docs.google.com/spreadsheets/d/1zMyTOp35qdsGZWCUxT74G337n8Z-Rj1FXpvdi4sve1c/edit#gid=0

Currently enabled: 332 functions

The functions found working in both debug and release mode were then enabled and the following testing was undertaken:

1) Created a new racer
2) Started a new tournament
3) Completed the first two races

Under debug mode, everything worked normally as expected. Under release mode, there was some intermittent crashing loading tracks, although functions worked correctly when independent. It appears release builds interact in strange ways (i.e. functions that simply call another function crash) that warrants further investigation

### DIFF
--- a/src/Dss/sithMessage.c
+++ b/src/Dss/sithMessage.c
@@ -10,7 +10,7 @@ int sithMessage_NetWrite(tSithMessage* pMsg, DPID idTo)
     HANG("TODO");
 }
 
-// 0x004049e0
+// 0x004049e0 HOOK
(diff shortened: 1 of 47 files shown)


tim-tim707:
Do you have an automated way of toggling function hooking ? Might be interesting to have

PR #18, opened 2024-03-27
Provide a way to cleanly call stdlib functions

DeonPoncini:
If you compile against mscvrt, the game will crash when using the cstdlib values from there, as the game has a statically compiled version of the C library included and the symbols clash

This provides a wrapper that be silently inserted (provide just an include) that maps back to these static functions

Start with just the functions within stdPlatform.c, but eventually expand to everything defined in swr_stdlib.h

### DIFF
--- a/src/Platform/cstdlib.c
+++ b/src/Platform/cstdlib.c
@@ -2,37 +2,144 @@
 
 // These are all c standard library functions. No decompilation necessary
 
-// stdio.h
+// ctype.h
+// 0x004a04e0
+long _filelength(int fd) {}
 
-// 0x0049f0f0 DO NOT DECOMPILE
+// 0x0049f440
+int _isctype(int c, int desc) {}
+
+// 0x0049f350
+int tolower(int c) {}
+
+// math.h
+// 0x0049ebf0
+void _fpmath(void) {}
+
+// 0x0049ed20
+long _ftol(double f) {}
+
+// stdlib.h
+// 0x0049ea60
+void _exit(int status) {}
+
+// 0x0049ef90
+double atof(const char* str) {}
+
+// 0x0049f040
+long atol(const char* str) {}
+
+// 0x0049f0e0
+int atoi(const char* str) {}
+
+// 0x0049fd80
+void* bsearch(const void* key, const void* base, size_t nmemb, size_t size, int (*compar)(const void *, const void *)) {}
+
+// 0x0049ea40
+void exit(int status) {}
+
+// 0x0049f200
+void free(void* ptr) {}
+
+// 0x0049f270
+void* malloc(size_t size) {}
+
+// 0x0049f8c0
+void qsort(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *)) {}
+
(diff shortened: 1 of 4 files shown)


tim-tim707:
Waiting for the extended version !

DeonPoncini:
This is now ready for final review. Please see my criteria used in the final commit for what I excluded out of the swr_stdlib.h header

tim-tim707:
Looks good to me but now swr_stdlib should be removed / merged into other files since it was supposed to be the current cstdlib. Right now we have swr_stdlib.h, cstdlib.c and Win95/std.c which can be confusing.

Merge conflict: I added this comment for `ftol()`:  swr_stdlib.h 279: void stdlib___ftol(); // https://github.com/NationalSecurityAgency/ghidra/issues/1246 . It looks to work nicely into ghidra now.

What do you advise here for swr_stdlib.h ?

DeonPoncini:
Thanks for the review linking that issue, I updated to head and pushed a new merge commit so it should be fine to merge now

Note: My interpretation of the folder setup was that Platform was setup as a abstraction layer, and the Win95 was for windows specific implementations - for example we could  have a Linux or Win11 directory or something with different implementations. As the C standard library is multiplatform already I inserted cstdlib in Platform. If the split is something else let me know

std.c/std.h: These are used for the standard library of the game itself. For example, things found like write_char/read_char that are neither the c standard or msvcrt should go here. Note: The wrapper functions fall into this, as stdFileSeek (at 0x0048c6c0) and fseek (at 0x004a0950) are different functions, and we should treat them differently. Perhaps given this, this should move from win95 to platform?

cstdlib.c/cstdlib.h: These are the entry point functions for everything found in the standard library, along with everything found in msvcrt (if you like I can create msvcrt.h/.c as well to separate out MS extensions, let me know). These are simply here to allow all the functions to work, and avoid crashes when dealing with some C functions (some work fine like sprintf, but others like fwrite crash, so might as well include them all). The idea is that this file will be deleted once decompilation is completed, replaced with just links to the c standard library instead. The way I wrote it specifically is so you can include this header, get the function inserted, then all you need to do is delete the header include and recompile and you are done, this is meant to be minimally invasive to just allow development to continue and the game running with the actual functions hooked

swr_stdlib.h - this is a superset of the two above, and contains things that I think are very useful for Ghidra, but not really for decompilation. Unless you think otherwise, I think any function that is only called from a C standard library function is not worth decompilation (just like I don't intend to decompile C standard functions either). This is because we aren't going to learn anything about the game, and they are easily replaced with a modern implementation of the C standard library once the game is fully decompiled. As such I think we should move this header out of src. It is very helpful to have documentation on all these internal functions so we can label them, but actually having them in the decompilation seems wasteful. Perhaps some separate directory that has Ghidra symbols? This is scripts/Ghidra today, as that is where master_functions.h and master_header.h live after generation, which are similar documentation headers that aren't used in the decompilation, just for symbol importing

tim-tim707:
Ok nice ! I will move swr_stdlib.h into archive/ then since its for documentation purposes but probably won't ever change

PR #19, opened 2024-03-29
Experiments2

tly000:


tim-tim707:
Great progress !
What is missing befote merge ?
Also, why are some parameters named `unk` instead of `node` when they come from `swrModel_node` ? Is it because decomp still looks weird ?

tly000:
> Great progress ! What is missing befote merge ? Also, why are some parameters named `unk` instead of `node` when they come from `swrModel_node` ? Is it because decomp still looks weird ?

i added some datasymbols just now, i think its ready now 🙂 

about the "unk", do you mean functions with `swrModel_unk*` parameters? those are function that operate on this `swrModel_unk_array` (where i said, that i think that those are viewports)

PR #20, opened 2024-04-06
Experiments2

tly000:
Some stuff in here is a bit more experimental and unfinished, you can decide if you want it in this state 🙂


src/Swr/swrRender.c
```diff
@@ -0,0 +1,207 @@
...
+void rdModel_Add3064NodeToScene2(swrModel_Node* node)
+{
+    HANG("TODO");
+}
+
+// 0x0044d1e0
+void ApplyD065PivotTransform(rdMatrix34* inout_transform, const rdMatrix34* transform, const rdVector3* pivot)
+{
+    HANG("TODO");
+}
```
tim-tim707: I was wondering if you could rename the `3064` and `D065` flags in the name of these functions and the  one below since it looks like you found what the flags were doing.

tim-tim707:
Sorry for the delay. It looks good overall but there is a little point I was intrigued about. It seems like you understood what the node flags where for (`5064`, `D064`, etc... ) but it seems like some function names weren't updated.
Amazing jobs on the models !

src/Swr/swrRender.c
tly000: fixed 👍

tim-tim707: Looks good to me !

tim-tim707:
Anything you want to update before I merge ?

tly000:
i have some more stuff but i think ill do another pull request then. thanks!

PR #29, opened 2025-01-25
Faster gltf loader

### LINKED ISSUE #27: Use a faster gltf parser

tim-tim707:
Fixes https://github.com/tim-tim707/SW_RACER_RE/issues/27
Its not much faster though, probably because of texture loading that should be progressive as well


tim-tim707 (commented):
👍

PR #33, opened 2025-06-29
Custom track fixes

tly000:
this is a workaround for the crash when loading custom tracks after vanilla tracks. the workaround is in swrObjJdge_InitTrack_delta. 

theres also another small fix for a bug in the load_replacement_if_missing function

### DIFF
--- a/dinput_hook/game_deltas/swrModel_delta.cpp
+++ b/dinput_hook/game_deltas/swrModel_delta.cpp
@@ -207,6 +207,11 @@ swrModel_Header *swrModel_LoadFromId_delta(MODELID id) {
         id,
     };
 
+    // setting this to 0 skips the generation of the renderDroid scene graph in the RenderAll
+    // functions. it's not needed since the renderer replacement uses the swrModel_Node scene graph #
+    // directly.
+    assetBufferModelLoaded = 0;
+
     return header;
 }
 
--- a/dinput_hook/renderer_hook.cpp
+++ b/dinput_hook/renderer_hook.cpp
@@ -1080,18 +1080,6 @@ extern "C" void init_renderer_hooks() {
                   (uint8_t *) swrSpline_LoadSplineById_ADDR);
     hook_replace(swrSpline_LoadSplineById, swrSpline_LoadSplineById_delta);
 
-    fprintf(hook_log, "Patching memory addresses\n");
-    fflush(hook_log);
-
-    patchMemoryAccess(0x0045b437 + 2,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, trackID));
-    patchMemoryAccess(0x0045b42d + 2,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, splineID));
-    patchMemoryAccess(0x0045b426 + 3,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, PlanetIdx));
-    patchMemoryAccess(0x0045b441 + 3,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, planetTrackNumber));
-
     fprintf(hook_log, "Done\n");
     fflush(hook_log);
 }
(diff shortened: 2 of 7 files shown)


dinput_hook/game_deltas/swrModel_delta.cpp
```diff
@@ -207,6 +207,11 @@ swrModel_Header *swrModel_LoadFromId_delta(MODELID id) {
         id,
     };
 
+    // setting this to 0 skips the generation of the renderDroid scene graph in the RenderAll
+    // functions. it's not needed since the renderer replacement uses the swrModel_Node scene graph #
+    // directly.
+    assetBufferModelLoaded = 0;
+
```
tim-tim707, line 214: Why is it needed to skip the generation of the scene graph then ?

dinput_hook/renderer_hook.cpp
```diff
@@ -1080,18 +1084,6 @@ extern "C" void init_renderer_hooks() {
...
-
-    patchMemoryAccess(0x0045b437 + 2,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, trackID));
-    patchMemoryAccess(0x0045b42d + 2,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, splineID));
-    patchMemoryAccess(0x0045b426 + 3,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, PlanetIdx));
-    patchMemoryAccess(0x0045b441 + 3,
-                      ((uint8_t *) g_aNewTrackInfos) + offsetof(TrackInfo, planetTrackNumber));
-
```
tim-tim707, line 1094: Why remove this ?

dinput_hook/replacements.cpp
```diff
@@ -604,7 +604,7 @@ void load_replacement_if_missing(MODELID model_id) {
 
         std::string filename;
         // Custom track, use folder to search for replacement model
-        if (currentCustomTrack.has_value() && model_id > CUSTOM_TRACK_MODELID_BEGIN) {
```
tim-tim707, line 607: If I recall correctly, there are case of custom tracks with id inferior to the CUSTOM_TRACK_MODELID_BEGIN value. I added the check in order not to overflow `modelid_cstr[]`. If this works properly, then you should also remove this part of the condition on lines 660 in the same function.

tim-tim707 (changes requested):
Great work overall ! As stated in a comment of yours this is still a hack / TODO for something better ? It'll probably become permanent anyways since I doubt there is going to be a lot of movement around these functions in the future.

I'm okay to merge this as soon as possible 👍. Thanks again for your contributions !

dinput_hook/game_deltas/swrModel_delta.cpp
tly000: its not needed to skip it, this just saves a bit of work. i can remove it if you want

dinput_hook/renderer_hook.cpp
tly000: this is already handled in the function patch_trackInfos_Usages 🙂

dinput_hook/replacements.cpp
tly000: there was a crash here, i think its because it should be model_id >= CUSTOM_TRACK_MODELID_BEGIN instead of >. 

but this check is not needed anyways, just currentCustomTrack.has_value() is enough. 

im a bit confused why it crashed there because model_id is not actually >= CUSTOM_TRACK_MODELID_BEGIN because the model_id parameter in this function comes from asset_pointer_to_model, which contains exactly `currentCustomTrack->model_id` (which in turn is in  the default model_id range, at least for the current custom tracks).

tly000:
> Great work overall ! As stated in a comment of yours this is still a hack / TODO for something better ? It'll probably become permanent anyways since I doubt there is going to be a lot of movement around these functions in the future.
> 
> I'm okay to merge this as soon as possible 👍. Thanks again for your contributions !

yes, there is some kind of bug that leads to invalid pointers in the scene graph. this bug can be detected by checking if the node type is a valid type. the invalid node has a child pointer with the adress 0xFFFFFF (at least on my machine).

dinput_hook/game_deltas/swrModel_delta.cpp
tim-tim707: Ok I'll keep it then

PR #34, opened 2025-06-30
disable glfw dinput replacement for now to make gamepads work correctly.

tly000:
i tried to add gamepad support to the stdControl dinput replacement functions. this did work, but the joystick mappings menu is broken and cannot be used properly. this is why i disabled the dinput replacement for now. 

this means that gamepads should work normally now.


tim-tim707 (approved):
LGTM !

PR #36, opened 2026-02-08
add texture replacement and msaa/aniso graphics settings

tly000:


tim-tim707:
Looks great overall.
Two small questions:
- What about the default framebuffer needing to be explicit ?
- Why skip pod textures ?

tly000:
> Looks great overall. Two small questions:

Thanks!
>     * What about the default framebuffer needing to be explicit ?

This way the MSAA settings can be changed while the game is running. changing the MSAA samples of the default framebuffer is not possible. this means this setting could only be changed when restarting the game. instead of rendering to the default window framebuffer, a new framebuffer is created as the "default" one. this new framebuffer can be changed freely to change the MSAA settings. at the end of swrViewport_Render_Hook this MSAA framebuffer is then copied into the window with glBlitFramebuffer. 

this means unbinding a framebuffer cannot use glBindFrameBuffer(0) because it would not revert to the MSAA framebuffer but to the non-msaa window framebuffer. 

> 
>     * Why skip pod textures ?

this is to help the creation of a texture pack. the pod textures are often mixed up with the map textures in the list and can be easily filtered this way. filtering/grouping the textures by model id would also be a nice option.

dinput_hook/game_deltas/DirectX_delta.c
```diff
@@ -82,12 +89,7 @@ void DirectDraw_LockZBuffer_delta(uint32_t *bytes_per_depth_value, LONG *pitch,
 }
 
 // 0x00431cd0
-void DirectDraw_UnlockZBuffer_delta(void) {
-    if (depth_data)
-        free(depth_data);
-
-    depth_data = NULL;
-}
```
tim-tim707, line 90: Where is depth_data freed if not here ?

tly000: its only freed, if the depth data size is changed, directly in DirectDraw_LockZBuffer_delta. this way the update is a bit faster because malloc/free is only needed if the window is resized.

tim-tim707: I think it should also be freed on program exit somehow just for good measure

tly000: i think the easiest way to do that is by making this source file a .cpp file and using a global std::vector. this will be cleanup up automatically in some cases. 

in practice this does not make much sense in my opinion. the cleanup only happens on normal "exit" calls and not on program crashes. the game does not use "exit" and currently crashes on exit anyways (if i remember correctly). the os will clean up the memory afterwards.

tim-tim707 (approved):
LGTM

PR #39, opened 2026-06-02
Document weather/HUD/viewport subsystems

louriccia:
## Summary

Adds documentation headers for three subsystems mapped during a weather/HUD reverse-engineering investigation:

- **`src/Swr/swrWeather.h`** (new) — particle effects (rain/snow). Includes a three-layer analysis of "weather doesn't render on modern displays" and a verified 8-byte EXE patch that fixes the primary cause (Layer 2: a hard-coded `-1000.0f` off-screen sentinel in `swrViewport_ProjectToScreen` that immortalises particles once `screen_width >= 1000`).
- **`src/Swr/swrPlayerHUD.h`** (new) — per-player split-screen HUD/overlay orchestrator: render loop, target indicators, rearview, world sprites, weather, distance text. Also documents the back-buffer occlusion sampler that silently skips its per-pixel test on modern 24/32-bit surfaces.
- **`src/Swr/swrViewport.h`** (+8) — adds `swrViewport_ProjectToScreen` (the `-1000.0f` sentinel write site, world-to-screen projection used by every HUD/weather renderer).
- **`src/swr.h`** (+4) — adds `decompressLZSS` (called by `swrModel_LoadFromId`).

Header-only PR — `_ADDR` defines, declarations, and documentation comments. No `.c` implementations and no behavioral changes.

## Verified weather-render bug fix

- Reproduced "no snow at 1920x1080" empirically. Particles spawn for one frame and then never respawn because the despawn condition `local_3c < -screen_width` cannot fire once the off-screen sentinel `-1000.0f` is no longer strictly less than `-screen_width`.
- Earlier dgVoodoo2-as-cause hypothesis refuted under DDrawCompat: both wrappers show the exact same `< 1000` threshold, so the bug is in the EXE, not the wrapper.
- **Patch:** rewrite the sentinel from `0xC47A0000` (`-1000.0f`) to `0xFF800000` (`-INF`) at `0x0042B86A` and `0x0042B881`. 8 bytes total. Tested on Howler Gorge and Andobi Mountain Run at 1920x1080.
- Full RCA, byte table, remaining Layer 3 artefacts (rain flicker, snow trails during sharp turns), and proposed follow-up work all live in [`src/Swr/swrWeather.h`](src/Swr/swrWeather.h).

### DIFF
--- a/src/Swr/swrPlayerHUD.h
+++ b/src/Swr/swrPlayerHUD.h
@@ -7,7 +7,7 @@
 //
 // Call graph per frame:
 //   swrPlayerHUD_RenderAllViewports (the per-frame render loop, from FUN_00445980)
-//     -> swrViewport_Activate / swrViewport_Setup  (per viewport, see swrViewport.h)
+//     -> swrViewport_SetCurrent / swrViewport_UpdateViewTransforms  (per viewport, see swrViewport.h)
 //     -> swrPlayerHUD_RenderViewport(viewport, secondaryPass)
 //          if (DAT_004b94c0)  // HUD orchestrator enable, set by swrPlayerHUD_Enable
 //              -> FUN_0042c1a0            (target indicators -- not yet named)
@@ -47,7 +47,7 @@
 
 // swrPlayerHUD_SampleOcclusion is called once per frame at the start of the
 // render phase (from FUN_00445980, before the per-viewport render loop). It
-// locks the back buffer via DirectDraw_LockMainSurface and samples raw pixel
+// locks the back buffer via DirectDraw_LockZBuffer and samples raw pixel
 // data around every HUD marker's projected screen position to detect
 // occlusion by world geometry. The resulting "coverage" counts are stored at
 // DAT_00e9a3c0 (target indicators), DAT_00e9a7e0 (distance text labels),
--- a/src/Swr/swrPlayerHUD.h
+++ b/src/Swr/swrPlayerHUD.h
@@ -1,0 +1,93 @@
+#ifndef SWRPLAYERHUD_H
+#define SWRPLAYERHUD_H
+
+#include "types.h"
+
+// Per-player split-screen HUD/overlay subsystem.
+//
+// Call graph per frame:
+//   swrPlayerHUD_RenderAllViewports (the per-frame render loop, from FUN_00445980)
+//     -> swrViewport_Activate / swrViewport_Setup  (per viewport, see swrViewport.h)
+//     -> swrPlayerHUD_RenderViewport(viewport, secondaryPass)
+//          if (DAT_004b94c0)  // HUD orchestrator enable, set by swrPlayerHUD_Enable
+//              -> FUN_0042c1a0            (target indicators -- not yet named)
+//              -> swrPlayerHUD_RenderRearview
+//              -> swrPlayerHUD_RenderWorldSprites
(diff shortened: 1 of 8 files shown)


src/Swr/swrWeather.h
```diff
@@ -0,0 +1,214 @@
...
+//     if (local_3c < (float)-DAT_00ec86c4 || ...)   // < -screen_width
+//         state = 0;
+//
+//   With the sentinel held at exactly -1000.0f:
+//     screen_width = 999   -> -1000.0 < -999.0  TRUE  -> despawn fires
+//     screen_width = 1000  -> -1000.0 < -1000.0 FALSE -> particles immortal
+//     screen_width = 1080+ -> -1000.0 > -1080.0 FALSE -> particles immortal
+//
+//   That is the exact 999/1000 threshold, originating from a developer assuming
+//   in 1999 that no screen would ever be >= 1000 pixels wide.
```
tim-tim707, line 98: Assumptions in the code that manifest in the missing particles is kinda funny

tim-tim707 (approved):
Looks good, I just need to integrate this into my ghidra version

src/swr.h
```diff
@@ -13,6 +13,8 @@
 
 #define swr_noop3_ADDR (0x00483ba0)
 
+#define decompressLZSS_ADDR (0x0042d520)
```
tim-tim707: Already defined in swrLoader.h as swrLoader_DecompressData

tim-tim707: If you could remove this one and rename swrLoader_DecompressData to swrLoader_DecompressLZSS please

src/Swr/swrPlayerHUD.h
```diff
@@ -0,0 +1,93 @@
...
+//     the per-track HUD elements AND configures weather (swrWeather_Enable/SetColor/
+//     SetVelocity/SetParticleCap/SetStretchFactor) based on the track/weather scenario.
+//
+// swrPlayerHUD_SampleOcclusion runs once per frame before the loop (also from
+// FUN_00445980); it CPU-reads the back buffer for HUD-marker occlusion -- see the
+// high-resolution caveat below and the fuller analysis in swrWeather.h.
+
+#define swrPlayerHUD_RenderDistanceText_ADDR (0x0042c510)
+
+#define swrPlayerHUD_RenderRearview_ADDR (0x0042c800)
```
tim-tim707: I have this function defined as UpdateLightStreakSprites on my end

tim-tim707 (changes requested):
Except some comments, everything matches what I have on my side.

Can you provide some global names for the arrays, and variables used by swrWeather please ?

src/Swr/swrPlayerHUD.h
louriccia: Same dup pattern turns up on a few more addresses:
`0x42c510`, `0x42cb00`, `0x42cca0`, `0x42d490`, `0x483cb0`. 

The descriptive names here collide with the `UpdatePlayerPositionSprites` / `UpdateUnknownIngameSprites1`/`2` / `UpdateIngameSprites` / `RenderAll` placeholders in swrModel.h/swrViewport.h. Harmless for now (these headers aren't `#include`d anywhere), but the new names read better than the `UpdateUnknown*` ones, so I'd lean toward renaming the swrModel side to match and dropping the placeholders.

tim-tim707 (approved):

PR #40, opened 2026-06-03
Widescreen UI (phase 1): un-stretch 2D sprites and text on non-4:3 framebuffers

louriccia:
## Summary
On non-4:3 framebuffers the 2D UI/sprite/text layer was stretched horizontally:
it scaled X by `screenWidth/640` and Y by `screenHeight/480` independently. The
3D world is unaffected (it's projected with the true aspect ratio). This fixes
the 2D **scale** so the UI renders at correct 4:3 proportions.

## Fix (phase 1)
- **Sprites** — hook `swrSprite_GetUIScale` (0x0044f640) to mirror the game's own
  vertical scale onto X (`xscale = yscale = screenHeight * recipH`). Note the
  design reciprocals at 0x004acce0/0x004accd8 are **doubles** (`FMUL qword`), and
  the screen dims at 0x00ec85e8/0x00ec86c4 are ints (`FILD dword`).
- **Text** — glyphs scale inside `rdProcEntry_Add2DQuad2` (0x0042d990) via their
  own double reciprocals (0x004ac630/0x004ac628, read only there). Patch the X
  reciprocal to match the Y scale (the constants are read-only `.rdata`, so a
  one-time `VirtualProtect` is needed before writing).

Implemented as `dinput_hook/game_deltas/swrSprite_delta.{c,h}`, registered in
`renderer_hook.cpp`.

## Verified in-game
Sprites and text render full-height and un-stretched at 2560×1377 (16:9).

## Known limitations / follow-ups (Mode B)
The UI is currently **left-anchored** (no pillarbox centering yet), and these
2D-overlay subsystems still use the old stretched X scale and need the same
uniform-scale + centering treatment: cursor hit-test, menu/list layout, lens
flare, weather, HUD/AI-racer text. Rendering and hit-testing are decoupled
(hit-test is CPU-side), so each needs the transform applied.

## Also included
- `ImportHeaderInfos.py` now falls back to name-only when a signature can't be
  parsed instead of aborting the whole import.
- Documents the swrUI menu system in `swrUI.h` + 2D-scale globals in
  `data_symbols.syms`: page stack/push/pop, `NewWindow/NewLabel/NewButton`,
  `AddSprite`, `HitTest`, `ProcessMouse`, `DefaultElementProc`, the F2 page
  procs, `swrSprite_GetUIScale`, `swrRender_InitScene`, etc.


louriccia:
This should probably just be an optional thing toggled with the debugger

scripts/Ghidra/ImportHeaderInfos.py
tim-tim707, line 1: Can you move this in a separate PR please ?

tim-tim707 (changes requested):
Looks good, can you add a checkbox in imgui debug menu to toggle the fix ?

Also move the changes to the import script in another PR so I can accept them separately

dinput_hook/game_deltas/swrSprite_delta.h
```diff
@@ -0,0 +1,12 @@
+#ifndef SWRSPRITE_DELTA_H
+#define SWRSPRITE_DELTA_H
+
+// 0x0044f640 swrSprite_GetUIScale
+// Widescreen UI fix (phase 1): replace independent X/Y screen scaling with a
+// uniform scale derived from the vertical axis, removing the 4:3 horizontal
+// stretch on non-4:3 framebuffers. Gated by imgui_state.widescreen_ui (toggle
+// in "graphics settings"); when off the original stretched behavior is
+// restored. See ghidra_analysis/ui_system_notes.md.
```
tim-tim707: Seems the `ui_system_notes.md` is not part of this PR.

I'd prefer to remove this comment but if you find value in adding the notes I'm ok to merge them as well

tim-tim707 (changes requested):
Except a reference about a missing file, I'm good to merge this

PR #42, opened 2026-06-03
swrSound: map the audio subsystem (cache, playback, mixer)

louriccia:
Name 25 functions of the A3D-based audio subsystem, with _ADDR defines and documented prototypes in swrSound.h. The subsystem: register sounds by name, load their audio on demand within a byte budget (with LRU/clock eviction), resolve and play them 3D-positionally, and mix across 8 channels each frame.

Resource cache: AllocBank, FreeBank, RegisterSound, LoadSoundMap, LoadDefaultBank, LoadSound, UnloadSound, ReadIntoSource, EvictSounds. Manager: Teardown, SetDefaultConfig, GetEntry, AcquireSource, UnloadAll. Playback/SFX: ResolveSfxId, PlaySpatial, PlaySpatialRange, PreloadSfx, PreloadRacerSounds, PreloadSoundSet, UpdateEngineAudio. Channel mixer: Update (the per-frame tick), ResetChannel, ResetChannels, RewindChannels.

The 0x4c-byte bank descriptor layout is documented inline. Source/descriptor pointers are typed void* for now (no swrSoundEntry struct yet). No behavior change; header-only.

PR #56, opened 2026-06-07
swrSound: reimplement the audio subsystem (16 functions)

louriccia:
Reimplements the `swrSound` subsystem in `src/Swr/swrSound.c` from the disassembly. After this, the only remaining stub in the file is `swrSound_Startup`, which is blocked on the cache layer from #42 (`LoadSoundMap` / `ReadAudioConfig` / `stdHashtbl_New`).

### Functions (16)
- **WAV load pipeline:** `ParseWave` (RIFF/WAVE PCM header), `NewSource` (builds the `WAVEFORMATEX`, `IA3d4::NewSource`), `CreateSourceFromFile` (open `data\wavs\{22K,11K}\<name>` → parse → alloc → stream PCM in)
- **Playback / COM:** `GetRenderMode`, `Play`, `GetWavePosition`
- **Sound-name hashtable:** `Find` / `Add` / `Remove` (`stdHashtbl` wrappers)
- **A3D engine lifecycle:** `Init` (CoCreate `IA3d4`, caps, cooperative level, listener) / `Shutdown`
- **Background streaming thread:** `CreateThread`, `SetPlayEvent`, `ThreadRoutine`, `UpdateStreaming`, and the buffer-fill helper `FillStreamBuffer` (was `FUN_004233a0`)

### Bug fixes / corrections
- `GetRenderMode` returned `(int)source` instead of the render-mode value — fixed (also repairs `SetGain`, which gates `Main_sound_gain_adjust` on `renderMode & 0x20`).
- `Play` and `GetWavePosition` were declared 1-arg, but the disassembly + every call site show they are `__cdecl` 2-arg: `Play(source, loop)` and `GetWavePosition(source, DWORD* out_pos)`. Prototypes corrected.
- `CreateSourceFromFile`: the original released an uninitialised stack slot on a parse failure and leaked the file handle on a `WriteLocked` failure — both fixed.

### Symbols added to `data_symbols.syms`
`swrSoundStream_file/loop/bytesRemaining/bufferSize/entry`, `swrSound_loadedBytes`, and `IA3dGeom_ptr` (the optional A3D geometry interface released at teardown; never instantiated by the game).

### Verification
- Builds clean (WinLibs GCC 13.2).
- **Runtime-verified**: temporarily forward-hooked all 16 (`// <addr> HOOK`) and ran the game — `hook_dinput.log` showed 16 `[Replace]` entries, SFX were correct, and streamed music played smoothly and looped cleanly. The HOOK annotations were reverted afterward, so the committed state is reverse-hooked per project convention.

### Note for merge
`swrSound_UpdateStreaming_ADDR` + prototype are also added by #42; declarations are matched exactly, so the overlap is a trivial same-value redefinition rather than a divergent duplicate.


tim-tim707 (approved):

src/Swr/swrSound.h
```diff
@@ -53,6 +55,8 @@ int swrSound_Remove(char* name);
 int swrSound_CreateThread(void);
 int swrSound_TerminateThread(void);
 DWORD swrSound_ThreadRoutine(LPVOID lpThreadParameter);
+unsigned int swrSound_FillStreamBuffer(void* entry, unsigned int writeCursor, unsigned int nbBytes);
```
tim-tim707, line 58: I have 4 arguments to this function and not 3, all used. Any idea ?

louriccia: Will look into it

louriccia: Looked into 0x4233a0 -- it's the per-frame stream refill (queries the A3D buffer, zero-fills, reads the next ~0x15888 chunk from the file, wraps to the loop point at +0x3c/+0x28). My local Ghidra mis-analyzed it and dropped a pointer, so my prototype came out 3-arg and I can't cleanly recover your 4th from here. Could you paste your 4-arg signature? I'll match it exactly.

PR #58, opened 2026-06-08
Fix hang/cursor stutter when closing via the window X

louriccia:
Closing the game with the window **X** briefly hung the system and stuttered the cursor; closing via the in-game **Quit Game** was instant. The X path also silently skipped the profile save.

**Cause:** the GLFW window in `Window_Main_delta` bypasses the original `Window_msg_main_handler`, so its `WM_DESTROY -> Main_Shutdown()` path never ran. On X-close the loop just exited and the process unwound with no graceful teardown, leaving the OS to forcibly reclaim the GL context, sound, and the still-acquired input devices.

**Fix:** when the loop exits, do exactly what the in-game quit does — `Main_Shutdown()` (saves the profile, stops sound, releases input/display in order) then `ExitProcess(0)`. `ExitProcess` skips the ~1s graceful GL/GLFW + C++ static-destructor unwind that a plain `return` would trigger, so the X now closes as instantly as the menu quit.

`Main_Shutdown` is reverse-hooked, so the call reaches the original `0x004240d0` rather than the reimpl stub.

Tested locally: cursor stutter/hang gone, close is instant, profile saved.

### DIFF
--- a/dinput_hook/game_deltas/Window_delta.c
+++ b/dinput_hook/game_deltas/Window_delta.c
@@ -394,6 +394,18 @@ int Window_Main_delta(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR pCmdLin
 #endif
     }
 
+    // Closing via the window's X sets glfwWindowShouldClose and drops out of the loop here.
+    // The GLFW window bypasses the original Window_msg_main_handler, so its WM_DESTROY ->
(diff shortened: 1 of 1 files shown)


tim-tim707:
Also dynamic patching crashes the main on exit (stack is offset / busted and should be re-ajusted). This should be fixed sonner rather than later, but requires x32dbg stepping and checking to do properly

PR #50, opened 2026-06-04
swrObjJdge: map the track/level load pipeline

louriccia:
Maps how a race spins up its world — the helpers `swrObjJdge_InitTrack`
orchestrates. 8 functions named in `src/Swr/swrObj.h`. No behavioral changes.

### Pipeline (InitTrack: models → HUD → spline → grid → lights → AI → racers)
- `SpawnRacers` — loads every racer's pod model (LOD vs full) and spawns the
  field in random grid order
- `SpawnRacer` — allocates each `Test` pod, wires its models, calls `swrRace_Init`
- `GetSpawnTransform` — 4-3-4-3 starting-grid placement from the spline
- `SetupStartingGrid` — positions the grid nodes from the start-line transform
- `SetupTrackEnvironment` — planet/track → spline id, loads it, sets fog / clear
  color / start camera
- `LoadCommonModels`, `GetTrackModelRoot`, `GetCustomStartTransform`

Ties the previously-mapped systems together: this is the loader that feeds the
pod (`swrRace_Init`), camera, and race-manager.

PR #44, opened 2026-06-04
input: map the controller subsystem (bindings, polling, force-feedback)

louriccia:
## What

Maps the input/controller subsystem — **33 functions** named with `_ADDR` defines and
documented prototypes (header-only, no behavior change). Two commits: the
`swrControl`/`stdControl` binding+polling+FF layer, then the lower-level `cifr_`
Immersion force-feedback effect wrapper in a new `src/Platform/cifr.h`.

The full flow: **initialize** the input devices → **load/parse** key + axis bindings
(deadzone, sensitivity, axis flip) → **poll** them per frame into game actions →
drive **force-feedback** effects via the Immersion SDK.

## swrControl / stdControl (commit 1)

**Input core** (per-frame polling, atop the now-documented `swrControl_ProcessInputs`):
- `swrControl_AxisAsButton` — analog-axis → digital threshold test
- `swrControl_ClearDeviceState` — clear a device class's accumulated inputs
- `swrControl_PollAccept` / `swrControl_PollCancel` — unified accept/cancel polls
- `swrControl_SnapshotKeyboard` — 256-key snapshot + buffered-key (WndProc) drain

**Bindings / config:**
- `swrControl_Initialize` — bring up the control system (devices, FF config, mappings)
- `swrControl_ClearBindings` / `swrControl_SetDefaultMappings` / `swrControl_LoadMappings` —
  clear, default-populate, and config-file-parse the binding tables
- `swrControl_ParseKeyName` / `swrControl_ParseFunctionName` — config token parsers
- `swrControl_ApplyAxisConfig` — push parsed sensitivity/deadzone into the axis registration

**stdControl axis setters:** `stdControl_SetAxisDeadzone`, `stdControl_SetMouseRange`

**Force-feedback (high level):** `swrControl_LoadForceEffects`, `swrControl_PlayForceEffect`,
`swrControl_StopForceEffect`, `swrControl_StopAllForceEffects`, `swrControl_PrepareForceEffect`

The binding-table layout (0xc-byte `{flags, inputId, actionId}` entries, terminated by
`0xff`; separate joystick/mouse/keyboard tables) and the deadzone/sensitivity globals are
documented inline.

## cifr_ — Immersion force-feedback effects (commit 2)

New `src/Platform/cifr.h` documenting the wrapper over the Immersion IFORCE / TouchSense
SDK (the `_IFCreateEffects_12` / `_IFReleaseEffects_8` / `_IFLoadProjectFile_8` imports).
[...]

PR #41, opened 2026-06-03
spline: complete LoadSpline byteswap + map the spline runtime

louriccia:
Finish the swrSpline_LoadSpline reimplementation (the control-point endian-swap loop; swrSplineControlPoint already matched the on-disk layout). Map and name the spline runtime/baking engine: control points form a directed graph keyed by a per-node progress band, baked at load into an arc-length progress table + fixed-step tessellation, and followed at runtime by an embedded cursor evaluated into orientation matrices.

16 functions named with _ADDR defines + prototypes in swrSpline.h / swrObj.h: loader (LoadSpline, LoadSplineById); cursor traversal/eval (CursorInit, CursorSeek, CursorStep, CursorGetNode, CursorEvaluate, Interpolate, EvaluateToMatrix, EvaluateAtOffset, FindNodeByProgress, ProjectPoint); post-load baking (Build, BuildProgressTable, TraceProgress, ForEachSample, ResetNodeProgress); and swrObjJdge_InitSplineCursor.

PR #59, opened 2026-06-09
Resolve duplicate _ADDR/name definitions + add a dup-check script

louriccia:
Resolves the duplicate `_ADDR`/name defs you flagged across the mapping PRs, and adds the dup-check you suggested.

**Renames to your canonical names** — all header-only (these addresses have no reimpl/hook refs, so it's a pure doc change):
- swrSpline: `FindNodeByProgress` → `getControlPointIdFromProgress`, `CursorGetNode` → `getControlPoint` (#41)
- swrControl: `LoadMappings` → `stdConfFile_readAndApplyConf`, `ParseKeyName` → `stdConfig_getKeymap_id`, `AxisAsButton` → `stdControl_isAxisAboveDeadzone` (#44)
- swrSound: `ReadIntoSource` → `swrSound_writeFile` (#42)
- `swrObjJdge_LoadCommonModels` → `swrModel_LoadBeamAndSparkModels` (#50)

Also dropped a duplicate `swrSound_UpdateStreaming` define + prototype (it was listed twice).

**The dup-check ("input script"):**
- `ImportHeaderInfos.py` now warns when a header name would rename a function that already has a curated (non-`FUN_`) name in your DB — that's the case most of these slipped through, since the canonical names live in your DB, not the headers.
- `CheckHeaderDuplicates.py` is a standalone host-side scan for in-repo collisions (one address / two names, dup defines). `GenerateMasterHeader.py` runs it automatically after regenerating.

**One to double-check:** I had `0x422f00` reading wav data from file into the source buffer, but your name is `swrSound_writeFile` — went with yours, flagging in case the semantics want a second look.

**Left out (your call — not blind header edits):**
- `0x408640` — `swrUI_UpdateProgressBar` vs `DirectDraw_BlitProgressBar` (the renderer hooks the latter via `_delta`).
- `0x417090` — `swrSprite_FreeSprites` vs `FreeSpritesMaterials` (also has a latent material-cleanup bug, and `hook_generated.c` depends on the prototype).
- The scan also surfaces some `cstdlib.h` placeholder-name collisions (`crt_internal33`, etc.) — pre-existing, left alone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ master_functions.h
 
 # python
 .venv
+__pycache__/
+*.pyc
 # node
 env
 a.out
--- a/scripts/Ghidra/CheckHeaderDuplicates.py
+++ b/scripts/Ghidra/CheckHeaderDuplicates.py
@@ -1,0 +1,136 @@
+# Scans every header in src/ for duplicate _ADDR definitions and reports:
+#   - one address defined under more than one name (a parallel define that
+#     should reuse the existing canonical name)
+#   - one name defined at more than one address
+#   - the exact same name+address #define written more than once
+#
+# This is the check tim-tim707 asked for: mapping PRs sometimes give a fresh
+# name to an address that already has a canonical _ADDR in another header.
+# Importing such a header into Ghidra then creates a second label at that
+# address. Run this before regenerating master_header.h / opening a PR.
+#
+# Plain CPython (host side, no Ghidra). Usage:
+#   python scripts\Ghidra\CheckHeaderDuplicates.py          # warn + exit 1 on dups
+#   python scripts\Ghidra\CheckHeaderDuplicates.py --warn   # warn but exit 0
+
+import os
+import re
+import sys
+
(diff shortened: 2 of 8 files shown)


louriccia:
On `swrSound_writeFile` (0x422f00): the body locks the source buffer, reads the wav from file into it, then unlocks -- so it's really a file->source load (reads disk, writes the source). Kept your `writeFile` name; just flagging in case `ReadIntoSource` / `LoadIntoSource` reads truer. Your call.

tim-tim707:
OK for LoadIntoSource then

tim-tim707 (changes requested):
except the previous rename comment all the scripts and changes are good to go

tim-tim707:
Once this is merged, checked and properly synced with Ghidra, I'll re-compute the progress 👀

louriccia:
Done -- renamed `swrSound_writeFile` -> `swrSound_LoadIntoSource` (eeac57b). That was the only outstanding item, so this should be good to merge + sync.

PR #72, opened 2026-06-12
swrUI: name element procs + list/element-core + menu-core internals

louriccia:
swrUI mapping in the 0x411000-0x41c000 range, split into 3 commits:

1. **F1 element procs** + the draw / hit-test / checkbox / sprite-slot internals they call (fills the "F1 callback" stub at the top of swrUI.h)
2. **list-widget API** + element-core lifecycle / accessors
3. **menu page-stack** + page builders + element layout/state

All header-only (`_ADDR` defines + prototypes, no behavior change), ~68 functions. This consolidates what were PRs #73 and #74 into one. Happy to rename anything to match your DB.

### DIFF
--- a/src/Swr/swrUI.h
+++ b/src/Swr/swrUI.h
@@ -493,4 +493,23 @@ void swrUI_SetCaretActive(int active);
 void swrUI_SetCaretRect(int x, int y, int w, int h);
 void swrUI_ProcessPendingClose(void);
 
+// ---- Panels, list ops, text-entry editing, lifecycle (UI-range mapping) ----
+#define swrUI_Shutdown_ADDR (0x00412e40)              // UI teardown (free tree + sprite materials + hash table)
+#define swrUI_NewPanel_ADDR (0x0041b260)              // class 8: a 9-slice framed panel
+#define swrUI_BuildPanelFrame_ADDR (0x00419830)       // build the panel's 9-slice background sprites
+#define swrUI_SetSpriteRect_ADDR (0x004171a0)         // configure a sprite slot's dest rect + texture UV scale
+#define swrUI_HandleTextEntryKey_ADDR (0x00418120)    // text-entry edit handler (insert/erase/cursor/commit)
+#define swrUI_ClearListSelection_ADDR (0x0041b300)    // deselect all list items
+#define swrUI_SortListItems_ADDR (0x0041b4d0)         // reorder list items by their assigned index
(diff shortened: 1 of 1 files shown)


src/Swr/swrUI.h
tim-tim707, line 1: Please keep all the define macro at the top of the file, sorted by hexadecimal address incrementing downward

louriccia: fixed

tim-tim707:
Since there are a lot of functions for swrUI, I think it would be good to split it in multiple files, or do some sub-namespacing ?
What do you think ?

On another note, I think you have an updated view of the swrUI struct as well as an enum for highlight / selected / visible and such ? Could you add it here ?

tim-tim707:
> On another note, I think you have an updated view of the swrUI struct as well as an enum for highlight / selected / visible and such ? Could you add it here ?

Nevermind, saw PR #75

louriccia:
> Since there are a lot of functions for swrUI, I think it would be good to split it in multiple files, or do some sub-namespacing ? What do you think ?

here's the cleanest split according to claude:

0x410fd0–0x41b690 = the pure widget toolkit (~140 fns) — the real "swrUI"
0x401xxx–0x403xxx = menu/settings page procs (arguably belong near swrConfig)
0x43bxxx–0x440xxx + 0x457xxx = front-end glue/loaders — and notably most of these take swrObjHang*, not UI elements, so they're really hangar front-end code that's just swrUI_-prefixed.

Could leave it as one file and make the section headers more clear/sub-namespacing or actually split them apart into other files if it feels too big.

tim-tim707:
Let's make the sections clearer with sub-namespacing prefix now then, according to the ranges provided. I let you decide what to put as suffix of swrUI_X for the menus and the glue code

louriccia:
Went with sub-namespacing over splitting files. Pushed the menu procs as `swrUI_Menu_*` (Main / SettingsHub / VideoSettings / AudioSettings / ForceFeedback) here, since those were decl-only and stay contained in this PR.

The front-end glue (`HandleCircuits`, `TextMenu`, `DrawRecord`, the `Load*` loaders, etc.) I'll do as `swrUI_Front_*` in a separate PR -- those have bodies in swrUI.c plus callers in the dinput_hook deltas, so it's a repo-wide rename. Core toolkit stays `swrUI_*`.

PR #73, opened 2026-06-12
swrUI: name list-widget API + element-core internals

louriccia:
More swrUI mapping in the 0x411000-0x41c000 range -- 23 functions, header-only (`_ADDR` + prototypes, no behavior change):

- **list-widget item management**: add/find/select items, save + restore selection, the layout/scroll pass, color propagation
- **element lifecycle + accessors**: recursive free, unlink, find-by-class, value-text get/set, visibility, focus (UI4), sprite-slot flag

Independent of #72 (different region of swrUI.h, appended at the end). Happy to rename anything to match your DB.


louriccia:
Folded into #72 (consolidated the swrUI mapping into one PR with 3 commits).

PR #74, opened 2026-06-12
swrUI: name menu page-stack + page builders + element layout

louriccia:
More swrUI mapping in the 0x411000-0x41c000 range -- 17 functions, header-only (`_ADDR` + prototypes, no behavior change):

- **page stack**: get current page / depth, reset, is-prev-page
- **menu builders** (used by `swrUI_BuildMenuPages`): the standard Nav (Quit/Cancel/Back) / OK / Default / Restore buttons, plus the aux overlay windows
- **element layout/state**: reparent, edge-align, center on screen, enable/disable (gray), refresh root, broadcast-to-windows

Independent of #72 and #73 (separate region of swrUI.h). Happy to rename anything to match your DB.


louriccia:
Folded into #72 (consolidated the swrUI mapping into one PR with 3 commits).

PR #91, opened 2026-06-14
swrEvent: reimplement the entity event/registry subsystem (14 functions)

louriccia:
Reimplements `swrEvent` — the foundational entity registry every subsystem calls. It was `HANG("TODO")` stubs (plus the accessors had no body at all).

It walks `eventManagerMain` (a NULL-terminated array of `swrEventManager*`, one manager per entity type: Test/Toss/Trig/Hang/Jdge/Scen/Elmo/Smok/cMan) and drives the per-frame F0–F3 hooks + the F4 sub-event broadcast ('Load'/'Aloc'/'Free'/…) over each type's instance array, plus the id/index lookups and alloc/free/init.

- Registers `eventManagerMain` (0x4bfec0, typed **`swrEventManager**`** — corrects the old commented `[][9]` sketch; the disasm loads it as a pointer-to-pointer), the lookup-cache globals (`swrEvent_lastManager`/`swrEvent_lastIndex`), and `swrEvent_f0SkipMask` in `data_symbols.syms`.
- `swrEventManager` + `swrObj` structs were already in types.h.

**Verification:** compiles with the project's exact C flags (incl. `-Werror=incompatible-pointer-types` / `-implicit-function-declaration`); all 15 functions defined; spot-checked the generated asm vs the original (e.g. `GetEventCount`: load table → iterate manager pointers (+4) → return `count`@+8 — faithful match). I couldn't run a full DLL link from a fresh worktree (cmake's generated-source step is finicky on a clean configure, unrelated to this change) — worth a confirming build in your configured tree.

Note: `CallF4` reads f4's return (`== 2` ⇒ stop iterating), so it's cast to an int-returning fptr at the call site; types.h keeps `f4` as `void`. Flagging in case you'd prefer to change the `f4` type.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -118,7 +118,7 @@ lensFlareSpriteIds 0x004B94E0 int[16]
 
 currentSpriteColor 0x004BFA0C uint8_t[4]
 
-# eventManagerMain 0x004bfec0 extern swrEventManager[][9]; # TODO: in swrEvent.h for the moment
+eventManagerMain 0x004bfec0 swrEventManager**
 swrObjHang_unused_state 0x004bfec8 swrObjHang_STATE = -1
 swrObjHang_unused_unk 0x004bfecc int = -1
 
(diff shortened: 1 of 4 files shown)


tim-tim707:
> Note: CallF4 reads f4's return (== 2 ⇒ stop iterating), so it's cast to an int-returning fptr at the call site; types.h keeps f4 as void. Flagging in case you'd prefer to change the f4 type.

It should be switch to f4 returning an int then (or is there something else ?)

louriccia:
Done — switched `f4` to `int (*)(swrObj*, int*)` and dropped the cast in `CallF4`. The F4 implementers (`swrObjTest_F4` etc.) were already returning int, so it's just the struct field + the dead reference-table casts. Recompiles clean.

PR #95, opened 2026-06-14
fix(mp): restore racer selection after a race

louriccia:
Fixes the long-standing multiplayer bug where you can't change racer in Select Vehicle after a race -- the cursor freezes (non-host typically stuck on Anakin). It hits both the host and clients.

**Cause:** in an MP race the hangar keeps `num_local_players == 2`, so `swrRace_SelectVehicle` walks the local-player index `current_player_for_vehicle_selection` hotseat-style to player 1, reading its input via `FUN_00469c30(swrUI_localPlayersInputDownBitset[player])`. But on every machine the local controller's input is funneled only into slot 0 -- so the moment the menu steps to player 1 (a remote player) the cursor can't move and re-seeds to that player's racer.

**Fix:** in MP each machine drives exactly one local racer, so pin selection to the single local player (`current_player_for_vehicle_selection = 0`, `num_local_players = 1`) in a `swrObjHang_F0` delta. Confirm then exits cleanly without stepping to a remote player. SP / split-screen untouched (guarded by `multiplayer_enabled`).

Verified in-game on a 2-machine session.


tim-tim707:
You'll have to explain it to me a bit more. (Edit: in vocal can work fine)
I'll have to try / we could both try to see if it works properly

louriccia:
I tested this one locally with two machines and am able to reliably reproduce the bug in vanilla or modded mp. Would be down to test it out but this would be a huge win if it works for everyone. Down for a call sometime this week but I'm usually only available 6-10pm ET on weekdays.

tim-tim707:
That would be 12pm-4am France time which is a bit late for me ^^
I'm not sure I can be available this weekend but I may get holidays next week.

I'll try to catch you late one evening if I get the chance.

I don't want to wait a lot for this change, if you can test with other laggy players like you tested the 100 lap that could be great as well

PR #96, opened 2026-06-14
fix(mp): async DirectPlay sends to stop game-thread stalls under packet loss

louriccia:
Fixes the multiplayer lobby-wide stutter on lossy links.

**Cause (measured):** timing the game-thread network calls showed the synchronous `IDirectPlay4::Send` (via `stdComm_Send`) blocks the game thread ~31-63 ms per call under packet loss -- it waits on DirectPlay's retransmit/ACK cycle (the stall quantizes to the ~15.6 ms Windows timer tick). At the ~31 Hz state-broadcast rate that blows the frame budget, so the whole loop (local physics included) hitches.

**Fix:** reissue sends through `SendEx` (vtable+0xC4) with `DPSEND_ASYNC`, so DirectPlay's own worker thread absorbs the wait and the game thread returns immediately. The caller's reliability flag is preserved (guaranteed stays guaranteed). Also bounds the per-frame `swrMultiplayer_PumpPackets` receive drain so a bursty peer can't make the receiver chew the whole backlog in one frame.

Both are toggle-able via `SW_RACER_RE.ini` (`mp_async_send`, default on; `mp_packet_cap`, default 32).

Verified on a 2-machine session with induced packet loss -- the host no longer drops frames under lag.


tim-tim707:
This also looks like a great news.

I'd be interested in testing this as well just to be sure before merging but otherwise look great !

PR #97, opened 2026-06-15
feat: 100-lap races + 1hr+ race timer

louriccia:
Lifts the vanilla 5-lap cap (menu now goes to 125) and fixes the in-game timer maxing out at 50:00. All in the `dinput_hook` delta layer - no decomp/header changes.

What's going on:
- The real 5-lap limit isn't the menu - it's `swrObjJdge_F2` indexing a fixed 5-slot per-lap-time array (`swrScore::results_P1_Lap1..5`) by the current lap with no bounds check, so >5 laps walk off the end and corrupt the score struct (crash within ~6-9 laps). Fixed with a verified in-place byte patch that de-indexes those accesses.
- Per-lap times are reconstructed from the running total so best/worst/average still work for any count. `<=5`-lap races keep the exact vanilla results screen; `>5` shows a compact best/worst/avg/total summary (a per-lap list can't fit on screen).
- The 50:00 timer was a hard `3000.0s` clamp in `F2` - raised to 24h, and the time formatters now break out an hours field (`H:MM:SS`) past an hour. Identical to stock under an hour.
- Menu cap is 125 because `numLaps` is a signed char; the engine's int `num_laps` handles far more, so going higher just needs decoupling the menu value from the char.

Verified in-game: 100-lap race counts/finishes correctly, timer reads past 1hr, results screen looks right for both <=5 and >5.

### DIFF
--- a/dinput_hook/game_deltas/swrObjJdge_delta.cpp
+++ b/dinput_hook/game_deltas/swrObjJdge_delta.cpp
@@ -1,14 +1,22 @@
 #include <macros.h>
+#include <cstring>
+#include <cstdio>
 #include "swrObjJdge_delta.h"
 
 extern "C" {
 #include <Swr/swrObj.h>
+#include <Swr/swrRace.h>
+#include <Swr/swrText.h>
+#include <Swr/swrSprite.h>
 #include <globals.h>
 
 extern FILE* hook_log;
 }
 
 #include "../hook_helper.h"
+
+extern "C" void hook_function(const char *function_name, uint32_t original_address,
+                              uint8_t *hook_address);
 
 int fixup_invalid_node_ptrs(swrModel_Node *&node) {
     if (!node)
@@ -46,8 +54,11 @@ int fixup_invalid_node_ptrs(swrModel_Node *&node) {
 // TODO hack: this is a workaround for a crash when loading custom tracks. sometimes the scene graph
 //  contains nodes with invalid child pointers, this happens after playing a vanilla track and
 //  afterwards a custom track. can be reproduced playing "Spice Mine Run" and then "Bowsers Castle 1".
+static void reset_lap_tracking(swrScore *scores); // 100-lap support, defined below
+
 unsigned int swrObjJdge_InitTrack_delta(swrObjJdge *judge, swrScore *scores) {
     const unsigned int x = hook_call_original(swrObjJdge_InitTrack, judge, scores);
+    reset_lap_tracking(scores);
     const int num_removed_nodes = fixup_invalid_node_ptrs(swrViewport_array[0].model_root_node);
     if (num_removed_nodes != 0)
     {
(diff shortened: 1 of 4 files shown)


tim-tim707:
Did you really play a 100-lap race ? I hope you turned on auto-pilot 😄

louriccia:
> Did you really play a 100-lap race ? I hope you turned on auto-pilot 😄

haha no but dom and dredge did lol. I did leave it open to test the >1hr game timer.

dinput_hook/game_deltas/swrObjJdge_delta.cpp
```diff
@@ -56,3 +67,376 @@ unsigned int swrObjJdge_InitTrack_delta(swrObjJdge *judge, swrScore *scores) {
...
+// because the swrText declarations + _ADDR macros are already C-linkage-visible in this file.
+// Must run during hook registration, before init_hooks() applies the detours.
+void swrObjJdge_RegisterTimeFormatHooks() {
+    hook_function("swrText_CreateTimeEntry", (uint32_t) swrText_CreateTimeEntry,
+                  (uint8_t *) swrText_CreateTimeEntry_ADDR);
+    hook_replace(swrText_CreateTimeEntry, swrText_CreateTimeEntry_delta);
+    hook_function("swrText_CreateTimeEntryPrecise", (uint32_t) swrText_CreateTimeEntryPrecise,
+                  (uint8_t *) swrText_CreateTimeEntryPrecise_ADDR);
+    hook_replace(swrText_CreateTimeEntryPrecise, swrText_CreateTimeEntryPrecise_delta);
+}
```
tim-tim707: I'd prefer if this was hooked and replaced at the same place as everything else. I don't really get the issue about doing it before init_hooks

dinput_hook/game_deltas/swrObjJdge_delta.cpp
```diff
@@ -56,3 +67,376 @@ unsigned int swrObjJdge_InitTrack_delta(swrObjJdge *judge, swrScore *scores) {
...
+static int g_bestLapNum[LAP_RACER_MAX];
+static float g_worstLap[LAP_RACER_MAX];
+static int g_worstLapNum[LAP_RACER_MAX];
+static float g_lastLap[LAP_RACER_MAX];
+static float g_prevTotal[LAP_RACER_MAX];
+static int g_prevLap[LAP_RACER_MAX];
+static bool g_lapFinished[LAP_RACER_MAX];
+// First laps' times, kept so the <=5-lap vanilla results screen (which reads the 5-slot per-lap
+// array that the de-index collapsed) can be refilled with the correct per-lap times.
+static float g_lapTimes[LAP_RACER_MAX][8];
```
tim-tim707: why 8 ?

tim-tim707: That could be 5 originals + top 3 but its not clear

dinput_hook/game_deltas/swrObjJdge_delta.cpp
```diff
@@ -56,3 +67,376 @@ unsigned int swrObjJdge_InitTrack_delta(swrObjJdge *judge, swrScore *scores) {
...
+                *(float *) ((char *) score + SCORE_LAP1 + i * 4) = g_lapTimes[r][i];
+        }
+        hook_call_original(swrRace_InRaceEndStatistics, jdge, score);
+        return;
+    }
+
+    // >5 laps: hide the per-racer pod portrait sprites the vanilla screen would manage, then draw
+    // the summary. Times are formatted here (hour-aware) and drawn left-aligned so they sit in a
+    // clean column instead of right-aligning back over the labels.
+    for (int i = 0; i < 0x13; i++)
```
tim-tim707: why 19 ? There can be more than 19 racers in a race. Maybe its 19 max visible ? I wouldn't be that sure

tim-tim707 (commented):
Looks great overall but there are some weird stuff none the less

dinput_hook/game_deltas/swrObjJdge_delta.cpp
louriccia: Good call - dropped the wrapper and moved the two `hook_function`/`hook_replace` pairs inline into `init_renderer_hooks()`, right next to the F2/results hooks. The only reason it lived in the delta file was symbol visibility; `renderer_hook.cpp` just needed `#include <Swr/swrText.h>`. There was no real ordering subtlety - that comment was misleading, so I removed it. Fixed in 9257266.

dinput_hook/game_deltas/swrObjJdge_delta.cpp
louriccia: No hidden meaning - it was just arbitrary headroom. That buffer only refills the vanilla <=5-lap results screen (which has 5 per-lap rows) and is never read on the >5-lap summary path, so it never needs more than 5 slots. Renamed it to `VANILLA_RESULTS_LAPS` (= 5) and sized everything by it so it's explicit - not 5 + top 3. Fixed in 9257266.

dinput_hook/game_deltas/swrObjJdge_delta.cpp
louriccia: You're right to question it - it's not racers, it's UI sprites. The original `swrRace_InRaceEndStatistics` single-player branch hides sprite IDs 0..0x12 (`while (i < 0x13)`), the per-lap row sprites it would otherwise lay out; the 2-player splitscreen branches instead clear 0xf..0x12 / 0x13..0x16. The >5-lap summary only runs single-player, so I mirror that exact range. Pulled the magic 0x13 into a named `kResultsScreenSprites` with a comment citing the original. Fixed in 9257266.

PR #100, opened 2026-06-15
stdConffile: reimplement the config-file parser

louriccia:
Reimplements the conffile parser/tokenizer that sits beneath `swrConfig` (and the various loaders that read `.cfg`/`.tab` files).

- **Open/Close** (read + write) -- read-open honours the `"none"` sentinel (in-memory context, no backing file) and pushes/pops the include stack when already open; write-open uses `"wb"`.
- **`ReadLine`** -- skips `;`/`#` comment lines, strips trailing `#` comments, lowercases, and handles trailing-backslash line continuation.
- **`ReadArgsFromStr`** / **`ReadArgs`** -- `strtok` key=value splitter (`", \t\n\r"` delimited) into `stdConffile_g_entry`.
- **`PushStack`/`PopStack`** -- the nested-include stack (saves/restores filename, line number, buffer, file handle, and the whole entry).

No new globals (all `stdConffile_*` state was already named). Verified with the project flags (both `-Werror` gates); disasm confirms the string literals (`none`/`wb`/`", \t\n\r"`), the `strtok`/`strchr`/`strlwr` calls, and the 4100-byte `g_entry` `rep movsl` copy in push/pop.

One small intentional divergence: `ReadArgsFromStr` caps at the `aArgs[512]` capacity; the original's pointer bound allowed one slot past the array (a harmless off-by-one never reached by real conf files).

### DIFF
--- a/src/General/stdConffile.c
+++ b/src/General/stdConffile.c
@@ -1,6 +1,9 @@
 #include "stdConffile.h"
 
+#include "globals.h"
 #include <macros.h>
+
+#include <string.h>
 
 // 0x004877b0 TODO: Crashes on release build, works on debug
 int stdConffile_Open(const char* pFilename)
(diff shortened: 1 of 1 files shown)


tim-tim707 (commented):
Looks good. Did you check that compilation is successful ?

PR #110, opened 2026-06-16
swrRace: reimplement the pod orientation / surface / ground-contact pipeline

louriccia:
Reimplements the pod orientation / surface / ground-contact pipeline in `swrRace` (the chain `swrObjTest_F2` runs each frame to keep the pod sitting on and aligned to the track):

- **`swrRace_InitUnk`** -> sets up and runs a ray-vs-collision-mesh query (was a stub).
- **`swrRace_ResetCollisionHit` / `swrRace_GetCollisionHit`** -> the hit-node result accessors.
- **`swrRace_RaycastGround`** -> casts "down" (surface-relative in magnet mode, with a straight-down retry) to find the ground; returns the contact distance + surface normal.
- **`swrRace_UpdateGroundContact`** -> the orchestrator: raycast, store the surface "up", slope-steer, gravity, track/wall collision, hover pads.
- **`swrRace_ApplySlopeSteering` / `...Magnet`** -> steer the pod along the ground slope; the magnet variant keeps it glued to a tagged surface.
- **`swrRace_ComputeTiltAngles` / `swrRace_AlignToSurface`** -> pitch/bank of the pod basis vs the surface, eased into the heading/tilt rates.
- **`swrRace_UpdateTurn2`** -> the per-frame orientation integrator (was a stub).

Names the collision-query globals it uses (`swrModel_collision*`, `swrRace_slopeAngle`, etc.).

Notes:
- These are reverse-hook reimpls (dormant; validated by compile + disasm), built on current master.
- The 9 not-yet-reimplemented callees (`swrModel_Collide*`, `swrSpline_EvaluateAtOffset`, `swrRace_{CollideBlockMove,DetectWallScrape,ApplyWallCollision,UpdateHoverPads}`) get HANG stubs so the dll still links.
- Full build links clean; duplicate-`_ADDR` scan clean (this caught + removed a `swrRace_ResetMatrixStack` that duplicated the canonical `rdMatrixStack44_Init`).
- Happy to match your DB names on anything.

### DIFF
--- a/src/Swr/swrRace.c
+++ b/src/Swr/swrRace.c
@@ -1,12 +1,15 @@
 #include "swrRace.h"
 
 #include "swrObj.h"
+#include "swrModel.h"
+#include "swrSpline.h"
 #include "swrEvent.h"
 #include "macros.h"
 #include "globals.h"
 
 #include <General/stdMath.h>
 #include <Primitives/rdVector.h>
+#include <Primitives/rdMatrix.h>
 
 // 0x00401340
 int swrRace_SelectProfileMenu(void* param_1, unsigned int param_2, unsigned int param_3, int param_4)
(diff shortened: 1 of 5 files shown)


src/Swr/swrRace.c
```diff
@@ -923,10 +976,117 @@ void swrRace_ActivateTriggersInRange(swrRace* a, swrModel_TriggerDescription* a2
...
+            *value = *value - swrRace_deltaTimeSecs * rate;
+            if (*value < target)
+                *value = target;
+        }
+    } else {
+        *value = *value + swrRace_deltaTimeSecs * rate;
+        if (target < *value)
+            *value = target;
+    }
+}
```
tim-tim707, line 993: Could it be this easy to patch the traction code by replacing deltatime with 1 / 30 sec ?

src/Swr/swrRace.c
```diff
@@ -1028,13 +1188,467 @@ void swrRace_ApplyGravity(swrRace* player, float* a, float b)
     a[2] += player->fallRate * gz;
 }
 
+// Normal-mode slope steering (no magnet). Projects world gravity (unk194_vec) onto the surface plane
+// to accumulate the downhill slide into velocitySlope (force quadratic in the steer term), and drives
+// unk8_1 (auto-tilt) from the downhill/facing alignment. Also publishes the slope angle to
+// swrRace_slopeAngle. Bails near-flat / near-inverted surfaces (dot(normal, worldDown) outside
+// [-0.995, 0.995]); out1 is scratch for the gradient, out2 the integrated slide.
+// 0x004791d0
```
tim-tim707, line 1196: update  unk194_vec to world_gravity then, as well as unk8_1 auto-tilt please

src/Swr/swrRace.c
```diff
@@ -1028,13 +1188,467 @@ void swrRace_ApplyGravity(swrRace* player, float* a, float b)
...
+        player->flags1 = flags1;
+
+        // surface-relative magnet mode: keep the "up" normal from tipping past ~horizontal/inverted
+        if (((flags1 & 0x400) != 0) && (up->z < 0.05f)) {
+            up->z = 0.05f;
+            rdVector_Normalize3Acc(up);
+        }
+        player->unk160.x = up->x;
+        player->unk160.y = up->y;
+        player->unk160.z = up->z;
```
tim-tim707, line 1481: is unk160 player up ?

src/Swr/swrRace.c
louriccia: swrRace_deltaTimeSecs is the per-tick delta and is ~1/30s at the target framerate, since the sim runs on a fixed logic tick -- so in normal play it's already effectively 1/30. I'd be cautious hardcoding the literal though: deltaTimeSecs is the single knob that carries any tick scaling (e.g. the FastMode cheat / slow-mo), so a literal 1/30 would desync those from vanilla. Worth tracing how IncrementFrameTimer computes it before swapping it in -- happy to do that as a follow-up.

src/Swr/swrRace.c
louriccia: Will do -- renaming unk194_vec -> world_gravity and unk8_1 -> autoTilt in a follow-up PR (they're swrRace struct fields used across the file, so it's a types.h rename + usages).

src/Swr/swrRace.c
louriccia: Yes -- unk160 is the live surface normal the pod treats as 'up' (RaycastGround writes the ground/collision normal into it each frame, clamped so it can't tip past vertical). I'll rename it to up in the same follow-up.

src/Swr/swrRace.c
louriccia: idk what cluade is on about but worth trying

PR #114, opened 2026-06-17
feat: gamepad rumble for modern controllers (XInput)

louriccia:
Modern controllers don't rumble in SWR: the DirectInput force-feedback path is gated on a legacy FFB device being detected at startup, which XInput pads never trip, so it's dead on modern hardware.

This adds an XInput rumble bridge in the `dinput_hook` delta layer. It reads the same `swrRace` state the game's own FF drivers use and drives the two XInput motors directly:

- collision / terrain shake (the physics "vibrator" value), L/R biased
- engine fire/damage, scaled by per-segment damage and biased to the damaged engine's side, with a repair-engage double-tap
- death jolt, boost-start afterburner (kick + sustain), wall-scrape sparks, and Sebulba's flamejet (right side)

Gated to active gameplay (Jdge entity present, not paused, valid pod), with a short post-unpause mute so quitting from the pause menu stays quiet. XInput is loaded dynamically (no new link dependency), and the low-frequency motor is boosted to feel balanced against the high-frequency one.

Debug menu gets a "Gamepad rumble" toggle + intensity slider (INI-persisted). Compile-time gated by `ENABLE_XINPUT_RUMBLE`. No `src` changes — entirely in `dinput_hook`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #115, opened 2026-06-17
feat: gamepad menu navigation (D-pad, START, BACK) for modern controllers

louriccia:
Modern XInput controllers already work for the stick + face buttons via the game's DirectInput layer, but the **D-pad, START, and BACK** were never mapped to anything. This adds a small XInput bridge in `dinput_hook` that reads them and feeds the game's *own* input paths — no binding-table changes.

What it does:
- **D-pad** navigates every menu — the swrUI page screens (splash/mode-select/options) via `swrUI_HandleKeyEvent` (the keyboard's own path), and the hangar front-end + in-race pause menu via the input bitsets they already read.
- **START** pauses/resumes a race and skips cutscenes (FMV movies + the in-engine cantina taunt scene).
- **BACK** cycles the HUD (the keyboard Caps-Lock function).

All in the delta layer (mostly one new `swrGamepadNav_delta.cpp`), toggleable in the ImGui settings (`enable_gamepad_nav`, INI-persisted) behind the `ENABLE_GAMEPAD_NAV` flag. Verified in-game. Sibling to the rumble PR #114 but independent (no shared files).

Along the way I named a few symbols that were `FUN_`/`DAT_` (`swrUI_UpdatePlayerMenuInput`, `swrUI_AccumulateMenuInputBit`, `swrControl_cancelPressedEdge`) — best-effort, happy to rename to match your DB.

### DIFF
--- a/dinput_hook/renderer_hook.cpp
+++ b/dinput_hook/renderer_hook.cpp
@@ -28,6 +28,7 @@ extern "C" {
 #include "./game_deltas/swrModel_delta.h"
 #include "./game_deltas/swrSpline_delta.h"
 #include "./game_deltas/swrObjJdge_delta.h"
+#include "./game_deltas/swrGamepadNav_delta.h"
 
 #include <glad/glad.h>
 #include <GLFW/glfw3.h>
@@ -951,6 +952,11 @@ extern "C" int stdDisplay_Update_Hook() {
     imgui_Update();// Added
     end_texture_replacement();
 
+#if ENABLE_GAMEPAD_NAV
+    // Latch the controller's D-pad / START / BACK state for the gamepad-nav hooks.
+    swrGamepadNav_Poll();
+#endif
+
     std::memset(replacedTries, 0, std::size(replacedTries));
     for (auto &[key, value]: additionnalReplacedTries) {
         value = 0;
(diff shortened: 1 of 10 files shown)


dinput_hook/renderer_hook.cpp
```diff
@@ -974,6 +980,11 @@ extern "C" void init_renderer_hooks() {
     // Hooks required for renderer replacement
     // ========================================
 
+#if ENABLE_GAMEPAD_NAV
+    // Feed the gamepad's D-pad / START / BACK into the game's menu + in-race input.
+    swrGamepadNav_RegisterHooks();
```
tim-tim707: Inline this function please

tim-tim707:
Looks good but there is some overlap with previous PR. This look like a stacked PR I'll have to check once the previous one is merged

louriccia:
Inlined -- registered the four hooks directly in `init_renderer_hooks` like the others and dropped the `swrGamepadNav_RegisterHooks` wrapper (delta fns now declared in the header).

On the overlap: this is independent of #114 (separate bridge, separate delta file) -- they just both touch the imgui-settings + CMake + renderer-hook init lines. Happy to rebase this onto master the moment #114 lands so it applies clean.

tim-tim707:
Tried playing the base game with controller and can confirm the controller doesn't work properly.
I want to try this but it needs a rebase / merge first

tim-tim707:
Works fine for the menus, with A and B for select / back.

However it is using the cross and neither the left or right thumbstick which is not intuitive. I suppose there is a way to map the thumbsticks instead of the cross ?

Also doesn't work to skip the race cinematic or control the pod. Is it planned for a next patch ?

I think this PR is good enough on its own to merge once the cross / thumbstick issue is sorted but I wonder if I should setup something else for proper xbox controller support

PR #116, opened 2026-06-17
swrSound: reimplement ResolveSfxId + the sfx flag/cooldown/recent leaves

louriccia:
Reimplements **`swrSound_ResolveSfxId`** (0x427110) and the sfx flag/cooldown/recent-play leaves it sits with.

`ResolveSfxId` maps a `(category, variant, id)` sfx request to a loaded sound's handle. Its tail had resisted decompilation (encrypted `.text` + an indirect jumptable) — I recovered it by reading the raw bytes and hand-disassembling. The logic:
- the per-category remap table only **validates** the id (a negative entry = no sfx for that id);
- the `.wav` filename is built from the category prefix (+ a *variant* prefix for categories 0-1) and the **raw id**, tried plain then with an `a` suffix;
- looked up via `swrSound_Find`, returning `found + 0x20` (the handle) or `-1`.

Every bounds check matches the prologue disassembly; register mapping confirmed (`esi`=id, `ebp`=category, `ebx`=variant, `edi`=prefix).

Also reimplements the clean leaves: `IsSfxOnCooldown`, `PushRecentSfx`, `WasSfxRecentlyPlayed`, `TestSfxFlag`, `SetSfxFlag`, `ClearSfxFlag`.

Names the tables/state these touch (8 per-category id-validity tables; category + variant name-prefix tables; the 3-entry recent-sfx ring + its write index; the per-category/variant cooldown timers) and renames the placeholder `unk_statuses` -> `swrSound_sfxFlags`. The cooldown threshold (`0x4ac47c` = 0.0) is inlined.

Build links clean; duplicate-`_ADDR` scan clean. One substitution: standard `sprintf` in place of the game's `stdlib__sprintf` (identical output; not declared in any header).

`MarkSfxPlayed` and the `PlaySfx*` functions are a follow-up — they share a get-sound-length helper (`GetEntry`→`field/1000`) and a category-2 cooldown-source table that aliases the recent-sfx ring, so they're cleaner done together. Happy to match your DB names on any of the new globals.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -106,6 +106,26 @@ swrSound_criticalSection 0x004b7e7e CRITICAL_SECTION
 
 swrModel_GlobalAnimationSpeed 0x004B7FA8 float
 
+# swrSound_ResolveSfxId per-category id-validity tables (id -> int16; a negative entry means
+# "no sfx for this id in this category"). Indexed by the raw id; the .wav filename is then built
+# from the category/variant prefix tables below + the raw id.
+swrSound_sfxRemap0 0x004b7fb0 int16_t[51]
+swrSound_sfxRemap1 0x004b8018 int16_t[38]
+swrSound_sfxRemap2 0x004b8068 int16_t[57]
+swrSound_sfxRemap3 0x004b80e0 int16_t[5]
+swrSound_sfxRemap4 0x004b80f0 int16_t[104]
+swrSound_sfxRemap5 0x004b81c0 int16_t[169]
+swrSound_sfxRemap6 0x004b8318 int16_t[105]
+swrSound_sfxRemap7 0x004b83f0 int16_t[168]
+
+# swrSound_ResolveSfxId sfx-name prefixes. Category prefixes (8: "ui","sp","raac","raig","rali",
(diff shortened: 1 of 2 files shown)


tim-tim707:
Looks good expect the merge conflict where two functions share the same address ?

```
// 0x004276a0
void swrSound_ClearSfxFlag(int index, unsigned int mask)
{
    swrSound_sfxFlags[index] &= ~mask;
=======
// 0x00427ad0
unsigned int swrSound_SelectPlanetIntroMusic(unsigned int planet)
```

louriccia:
Rebased onto master — the conflict was just both functions landing at the tail of `swrSound.c` (right after `GetHardwareFlags`), so I kept both: the `ResolveSfxId` block, then `SelectPlanetIntroMusic` after it. `data_symbols.syms` merged clean. Builds + links clean now and the PR shows mergeable.

Unrelated heads-up: master currently doesn't compile — `swrObj.c:320` still writes `jdge->unk1c4`, but that field (offset 0x1c4) got renamed to `aiSpeedSetting` in the AI-difficulty merge and the consumer wasn't updated. One-liner, sending a separate fix PR.

PR #120, opened 2026-06-17
swrRace: name the engine-fire FX + spline-cursor functions

louriccia:
Names five previously-`FUN_` per-pod race-tick functions from the decompilation, declaration-only (`_ADDR` + prototype), added as a themed group near the lap/frame block (swrRace.h is theme-grouped, not strictly address-sorted).

- `swrRace_UpdateFireEffects` (0x0047e450)
- `swrRace_InitFireEffects` (0x0047e580)
- `swrRace_AdvanceSplineCursor` (0x0047f8e0)
- `swrRace_UpdateSplineBinding` (0x0047fbb0)
- `swrRace_ComputeTrackOffset` (0x0047fca0)

`swrRace_DrawDebugPosition` (0x00483be0) is in the same range but decompiles unreliably (uninitialized stack), so I left it for later rather than guess its prototype.

### DIFF
--- a/src/Swr/swrRace.h
+++ b/src/Swr/swrRace.h
@@ -184,6 +184,13 @@
 
 #define swrRace_IncrementFrameTimer_ADDR (0x00480540)
 
+// engine-fire FX + spline-cursor track following (per-pod, run during the race tick)
+#define swrRace_UpdateFireEffects_ADDR (0x0047e450)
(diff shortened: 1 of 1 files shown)


tim-tim707:
Here is the prototype I have for the function at 0x00483be0 `void debug_ui_show_pos_comp(int debug_level)`

PR #125, opened 2026-06-18
renderer: reproduce the pod cable curve in the GL replacement

louriccia:
The cockpit->engine cables render flat/thin/unshaded in the GL renderer replacement. The game curves them each frame (`FUN_00481c30`) on the rd3d-converted mesh, which the replacement never builds — so it only ever drew the straight authored mesh.

This regenerates the game's curved cable in the GL path: small hooks on the per-pod cable updaters (`swrRace_PoddAnimateVariousThings` in-race, `FUN_004337e0` for hangar/menu/cutscene display pods) record each cable's bend amplitude, and the renderer rebuilds the cable from the game's own template (eased + sine-bent triangular tube with the baked vertex shading). Works in-race (incl. 4-engine pods like Ben Quadinaros), while paused, in the hangar "inspect vehicle", and the selection menu.

Pure `dinput_hook` delta layer — no `src`/header changes. `FUN_004337e0` is referenced by raw address (like the existing `std3D_*` raw-addr hooks); happy to name it in the DB if you'd prefer.


dinput_hook/renderer_hook.cpp
```diff
@@ -1162,6 +1251,17 @@ extern "C" void init_renderer_hooks() {
...
 
+    // Record each pod's cable-curve state per frame so the GL walk can bend the cables
+    // (the game's cable deformer only touches the rd3d mesh the replacement doesn't use).
+    hook_function("swrRace_PoddAnimateVariousThings",
+                  (uint32_t) swrRace_PoddAnimateVariousThings,
+                  (uint8_t *) swrRace_PoddAnimateVariousThings_ADDR);
+    hook_replace(swrRace_PoddAnimateVariousThings, swrRace_PoddAnimateVariousThings_delta);
+
+    // Display-pod animator (hangar inspect / selection menu / cutscenes) - register its cables too.
+    hook_function("swrRace_AnimateDisplayPod", (uint32_t) 0x004337e0,
```
tim-tim707: This function may be correct but its not yet in the decompilation. Can you add it in this PR please ?

louriccia: Done in f3f34d2 — named it `swrRace_AnimateDisplayPod` with an `_ADDR` define + prototype in `swrObj.h`, so the hook and the delta now reference the named address instead of raw `0x004337e0`. Kept it as a declaration rather than reimplementing the ~5800-byte body. One note: it lives in the `swrObjHang` address range but parallels `swrRace_PoddAnimateVariousThings`, so I kept the `swrRace_` prefix — easy to switch to `swrObjHang_` if you'd rather.

tim-tim707 (changes requested):
Still missing the swrRace_AnimateDisplayPod_ADDR decompilation define and function prototype

PR #133, opened 2026-06-19
imgui: live hook.log viewer window

louriccia:
The "Show logs" checkbox in the debug overlay displayed `imgui_state.logs`, a string nothing ever wrote to, so the panel was always empty.

Replaced it with a **Show Log** button that opens a separate window (imgui's `ExampleAppLog` style: Options / Clear / Copy / Filter / auto-scroll, plus light red/yellow tinting for error/warn lines). It tails `hook.log` -- the file every `fprintf(hook_log, ...)` already writes to -- through a private shared read handle, mirroring new bytes each frame. So it surfaces the real log stream with zero changes to the existing call sites.

Tailing only runs while the window is open, and scrollback is capped at 20k lines so a long session can't grow the buffer without bound.

Touches only `dinput_hook/imgui_utils.{cpp,h}`. Built + linked clean against current master.


dinput_hook/imgui_utils.cpp
```diff
@@ -299,6 +300,176 @@ void imgui_Update() {
...
+        for (const char *p = s; p + n <= e; p++)
+            if (strncmp(p, needle, n) == 0)
+                return true;
+        return false;
+    }
+
+    static bool line_color(const char *s, const char *e, ImVec4 &out) {
+        if (line_contains(s, e, "FAIL") || line_contains(s, e, "ailed") ||
+            line_contains(s, e, "rror") || line_contains(s, e, "Couldnt") ||
+            line_contains(s, e, "Aborting")) {
```
tim-tim707: Looks like its missing first letter on Failed and error

dinput_hook/imgui_utils.cpp
```diff
@@ -299,6 +300,176 @@ void imgui_Update() {
...
+            out = ImVec4(1.0f, 0.4f, 0.4f, 1.0f);
+            return true;
+        }
+        if (line_contains(s, e, "arn") || line_contains(s, e, "WARN")) {
```
tim-tim707: same with warn here

dinput_hook/imgui_utils.cpp
louriccia: This was claude's cheap way of being case insensitive. Updated to be more explicit about matching the whole word in that way.

tim-tim707 (approved):

PR #136, opened 2026-06-20
swrRace: use swrVehicleReaction enum instead of raw hex flags

louriccia:
You mentioned preferring that magic bit flags go through a named enum in `types_enums.h` rather than raw hex. Starting on that here.

`swrRace_UpdateSurfaceTag` tested `behavior->vehicle_reaction` against bare hex masks (`& 0x2000`, `& 1`, ...). All 13 map exactly onto the existing `swrVehicleReaction` enum, so I swapped them (Lava/Fall/Flat/Soft/ZOn/ZOff/Fast/Slow/Swst/Slip/Swmp/Mirr/Side) and retyped the field `uint32_t` -> `swrVehicleReaction`. The enum is int-backed, so the packed `swrModel_Behavior` layout is unchanged; the only other writer is the `SWAP32` byteswap in `swrModel.c`.

Built + linked clean. There are more raw-flag spots around (swrObjTest `flags0`/`flags1` especially), but those bits aren't named yet, so I'd rather do the RE first and send them as a follow-up.


tim-tim707:
Can you rename swrVehicleReaction enums by prefixing them please ?

louriccia:
Done - prefixed all 19 with `swrVehicleReaction_` (e.g. `swrVehicleReaction_Lava`) and updated the consumers. Also merged current master in so it links clean. Full `dinput.dll` build is green.

PR #139, opened 2026-06-20
Ci test

tim-tim707:


tim-tim707:
Compilation check works fine, with pinned version from the readme

PR #140, opened 2026-06-20
Bind HD pod replacement to the owning racer entity

louriccia:
﻿The HD pod replacer drew every pod from the single global `currentPlayer_Test`. With `ai_full_lod` on (every AI becomes a full `_pod`), each AI pod hit the player draw path and got stamped on the local player - a pile of HD pods riding you. Latent with one replacement model; shows up with a full set.

This resolves each pod node to its owning racer via the `swrScores[]` roster (each racer loads a distinct pod copy and keeps its own pod node in `unk1994_node`) and draws from that entity's own `engineXf`/`cockpitXf`. The per-MODELID `replacedTries` dedup is dropped for pods (the traversal visits each pod node once), so N racers draw at their own positions; the respawn blue-flash tint is per-entity now, so AI pods flash too.

`try_replace_pod` is reduced to hangar inspection only, which drops the dead `currentPlayer_Test` in-race branch and a raw address literal; the env/track scene-root checks use the named `someRootNode`/`someUnkRootNode` globals.

Playtested on a 12-racer grid with full-LOD AI + multiple HD pods: each draws on its own pod, no pile. Hangar inspection still fine. Includes a small Pod node owners ImGui dev readout - easy to drop if you would rather not have it.

Separate known issue (not this PR): the Tatooine pod reflection drops out after ~5s - confirmed to be the game dropping the reflection-subtree visibility (vanilla GL breaks the same way), unrelated to this change.

### DIFF
--- a/dinput_hook/imgui_utils.cpp
+++ b/dinput_hook/imgui_utils.cpp
@@ -14,6 +14,7 @@
 
 #include "replacements.h"
 #include "renderer_utils.h"
+#include "node_utils.h"
 #include "texture_replacement.h"
 #include "backends/imgui_impl_glfw.h"
 #include "backends/imgui_impl_opengl3.h"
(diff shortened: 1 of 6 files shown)


louriccia:
Heads up: the red integrity-check here isn't from this PR. master is red too (Merge #138) with the same std3D_delta.cpp:197/321 cast error. The WinLibs PATH export sits inside the cache-miss-only Download WinLibs step, so on a cache hit the build falls back to the runner 64-bit g++ and the pre-existing (GLuint)pointer casts error out. Fix in #141. This branch links clean on 32-bit WinLibs locally and should go green once #141 lands and this re-runs.

tim-tim707:
image

PR #148, opened 2026-06-21
swrWeather: declare the particle-pool sprite-id setter + reset

louriccia:
Declares 2 swrWeather functions named in the Ghidra DB but not in `src/` yet -- they complete the particle-pool API already documented in `swrWeather.h`:

- `swrWeather_SetParticleSpriteId(int slot, int spriteId)` -- set a pool slot's swrSprite id
- `swrWeather_ResetParticles(void)` -- clear all slot states + set every sprite id to -1

Decl-only (`_ADDR` + prototype in `swrWeather.h`), no behavior change. `_ADDR` macros appended in address order after `swrWeather_Enable`. Builds clean (full `libdinput_hook.dll` link).

### DIFF
--- a/src/Swr/swrWeather.h
+++ b/src/Swr/swrWeather.h
@@ -207,6 +207,10 @@
 
 #define swrWeather_Enable_ADDR (0x0042d450)
 
+#define swrWeather_SetParticleSpriteId_ADDR (0x0042d460)
+
+#define swrWeather_ResetParticles_ADDR (0x0042d470)
+
 void swrWeather_RenderParticles(void* viewport);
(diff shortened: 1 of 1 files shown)


src/Swr/swrWeather.h
```diff
@@ -207,6 +207,10 @@
 
 #define swrWeather_Enable_ADDR (0x0042d450)
 
+#define swrWeather_SetParticleSpriteId_ADDR (0x0042d460)
```
tim-tim707, line 210: I have swrSprite_newFlare here

Issue #143, opened 2026-06-20
Multiple ideas


tim-tim707:
- [ ] hd renderer full toggle
- [x] fps limiter
- [ ] disable multiplayer pod collisions
- [x] toggle hd font
- [ ]  upgrades in multiplayer
- [x] 5+ laps multiplayer
- [ ] sound effect loop glitch
- [x] crash handler with stack dump 
- [ ] ui highlight bug (high priority)
- [ ] even harder ai
- [ ] ai in multiplayer
- [ ] Prevent crash on missing asset/ directory
- [x] Better organisation of the Imgui panel (with a dev tab)
- [ ] collision viewer in multiplayer
- [ ] skip cutscene with controller
- [ ] hd renderer lighting issues when selecting pod in multiplayer (cantina is tagged as replaced)
- [ ] game crash when retiring as last member of lobby
- [x] name instead of # place on top of pods when racing
- [x] # place are wobbly instead of being cleanly projected
- [ ] Regression: Renderer replacement pod doesn't hide destroyed engine when spining out of control. Doesn't read the node visibility flag properly anymore ?
- [ ] Rumble / Vibration profile (filter by event, select strength)
- [ ] slash commands in chat ? - ready flag for player
- [ ] replace vpn with embedded p2p library (iroh ?)

louriccia:
@
A few more of these are done and can be checked off:

- **disable multiplayer pod collisions** — #205 (merged)
- **upgrades in multiplayer** — #204 (merged)
- **sound effect loop glitch** — #232 (merged) — the shared splash voice was restarting the player's loop under full-LOD AI
- **skip cutscene with controller** — #115 (merged) — START skips FMVs + the in-engine cantina taunt

Partial (leaving unchecked): **Rumble / Vibration profile** — #114 adds the toggle + strength slider, but no per-event filtering yet (and still open).
@

tim-tim707:
Superseded by 277

PR #149, opened 2026-06-21
Add FPS limiter and FPS overlay

louriccia:
image

Adds an optional frame-rate cap and an FPS overlay to the HD renderer (both default off / unobtrusive).

- **Frame rate cap** (`FPS` accordion in the F5 menu): 10-200 FPS or Unlimited, with -/+ paddles and 24-60 quick-set buttons. Paces the GL present via a high-resolution waitable timer + short spin tail. Doubles as a stopgap for high-FPS physics quirks. Persists to the ini.
- **FPS overlay** (top-right): big outlined number + frame time, optional smoothed-FPS graph with a fixed scale and average line. Independent of the F5 debug menu.
- Only active on the OpenGL/HD renderer path (`RENDERER_REPLACEMENT=ON`).

Part of #143.

### DIFF
--- a/dinput_hook/imgui_utils.cpp
+++ b/dinput_hook/imgui_utils.cpp
@@ -80,6 +80,22 @@ void read_settings_ini() {
         imgui_state.anisotropy = anisotropy;
     }
 
+    imgui_state.target_fps =
+        GetPrivateProfileIntW(L"settings", L"target_fps", 0, ini_path.c_str());
+    if (imgui_state.target_fps != 0) {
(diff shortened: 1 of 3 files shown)


tim-tim707:
I think this needs a rebase. I cannot apply patch and its 50+ commits late from master, even when there is no merge conflict

louriccia:
Merged current `upstream/master` in (was ~78 behind). Only conflict was the new `show_pod_names` field landing right next to the FPS-overlay fields in `ImGuiState` — kept both. Builds + links clean against master, so it should apply now.

PR #150, opened 2026-06-21
Add HD font replacement toggle

louriccia:
Adds an "Enable HD fonts" toggle (F5 menu, under "Enable HD model replacement") so players can swap the game's built-in fonts for the HD replacement set. Defaults on (current behavior), persists to `SW_RACER_RE.ini` (`[settings] hd_font`). Closes the "toggle hd font" item on #143.

**Applies live, no restart** (per your "journaling is up, we can move with this" note). Built on the #209 patch journal:

- `swrText_InitFonts_delta` now always runs the original first, so the built-in font materials load and the font-table slots hold valid pointers.
- It then builds the HD materials once (kept resident) and, if enabled, `PatchPointer`s the HD materials over the 7 font-table slots tagged `"hd_font"`.
- The F5 checkbox flips it live: enable journals the swap, disable is `UndoOwner("hd_font")` which restores the built-ins. Texture conversion only ever runs at load, so the runtime toggle is a pure (atomic) pointer swap.
- The vanilla `swrText_InitFonts` and the HD path write identical font-metadata globals, so only the 7 material-slot pointers need journaling.
- If the HD font assets are missing, the toggle now fails gracefully and keeps the built-in fonts (no more `assert`).

While here: the delta was named `swrModel_LoadFonts_delta` but hooks `swrText_InitFonts` - renamed it to match, and swapped the raw `0x0042d720` literal in its registration for `swrText_InitFonts_ADDR`.

Rebased onto current master. Built + linked clean (WinLibs i686 GCC 13.2); the three integrity-check scripts pass locally. Playtested: default HD, toggle off -> vanilla fonts live, back on -> HD live, rapid toggling stable, normal boot unchanged.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -173,6 +173,10 @@ InRaceSpritesEnabled 0x004B94C0 int
 
 lensFlareSpriteIds 0x004B94E0 int[16]
 
+# The 5 bitmap fonts built by swrText_InitFonts, referenced through the 7-entry font table. Each
+# swrFont holds pageCount glyph-page swrMaterial*s at +0x08; the HD-font replacement swaps those.
+swrText_fonts 0x004bf7e0 swrFont[5]
+
 currentSpriteColor 0x004BFA0C uint8_t[4]
 
 # extended-character (code > 0x96) compose tables read by swrText_GetStringWidth/Height/DrawString.
--- a/dinput_hook/game_deltas/swrModel_delta.cpp
+++ b/dinput_hook/game_deltas/swrModel_delta.cpp
@@ -13,6 +13,7 @@
 #include "../custom_tracks.h"
 #include "../nv_dds/nv_dds.h"
 #include "../imgui_utils.h"
+#include "../patch.h"
 
 #include <regex>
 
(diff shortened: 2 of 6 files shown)


tim-tim707:
Since its restart to apply, let's wait on the config and modding api v2 to toggle it properly and then merge

tim-tim707:
Since journaling is up, we can move with this

louriccia:
Done - reworked onto the journal so it toggles live now (no restart). swrText_InitFonts always runs the original first so the built-in fonts are resident, then journals the HD materials over the 7 font-table slots ("hd_font" owner); the F5 toggle is PatchPointer on / UndoOwner off. Rebased on master, mergeable, integrity scripts pass, playtested (toggle off->vanilla, on->HD, all live and stable).

louriccia:
Pushed a follow-up that drops the raw address literals: the 5 fonts are the swrFont[5] table at 0x004bf7e0, so I named it swrText_fonts in data_symbols.syms and reference the slots as swrText_fonts[f].pages[p]; the converter + font loader now use their existing _ADDR macros. Builds clean, all three integrity scripts pass, addresses are provably unchanged (e.g. swrText_fonts[3].pages[0] == 0x4bf920).

PR #154, opened 2026-06-22
MP: player names above pods (+ overhead-labels toggle)

louriccia:
image

Closes part of #143 ("names above opponents").

In multiplayer, the small label the HUD already draws over each pod now shows the **human player's name** instead of the race-position number. Single-player position numbers over AI are untouched.

### How
Rather than re-implement the world->screen projection (and its ~13 tuning constants), it wraps `swrPlayerHUD_RenderDistanceText` and only swaps the string it draws:
- SP (`multiplayer_enabled == 0`): calls the original verbatim, numbers unchanged.
- MP: builds a per-slot name label (`swrScores[N]` index == player number -> `obj.id` slot -> `swrMultiplayer_GetPlayerNameAscii`), sets a redirect flag, runs the original, then clears it. While set, `swrText_CreateTextEntry2` maps each draw back to its slot by the screen position the renderer just stored and swaps in the name. Blank player name falls back to the character name.
- Labels render at half size via the `~F0` text code and are centred with `~c`; the draw position is scaled 200% to undo `~F`'s 0.5 position offset.

### Toggle
Adds an **"Overhead racer labels"** debug-menu checkbox (`show_pod_names`, persisted to `SW_RACER_RE.ini`) that gates the labels in both SP and MP.

### Cleanup
- Names the swrText half-scale flag `0x0050c0ac` -> `swrText_halfScale` in `data_symbols.syms`; the delta clears it (so the minimap text after the label batch isn't shrunk) via the named global.

Verified in-game on **two machines** (host + client): names show centred and half-size over the correct pods, SP numbers unaffected, minimap text unaffected.

### DIFF
--- a/dinput_hook/game_deltas/swrPlayerHUD_delta.cpp
+++ b/dinput_hook/game_deltas/swrPlayerHUD_delta.cpp
@@ -1,0 +1,185 @@
+#include "swrPlayerHUD_delta.h"
+
+#include <windows.h>
+#include <filesystem>
+#include <cstdio>
+#include <cstdlib>
+
+extern "C" {
+#include <macros.h>
+#include <Swr/swrPlayerHUD.h>
+#include <Swr/swrText.h>
+#include <Swr/swrMultiplayer.h>
+#include <globals.h>
+}
+
+#include "../hook_helper.h"
+#include "../imgui_utils.h" // imgui_state.show_pod_names (the debug-menu toggle)
+
+// ===========================================================================================
+// Multiplayer: player names above pods.
+//
+// The HUD already draws a small label above each racer (the race position number) via
+// swrPlayerHUD_RenderDistanceText -> swrText_CreateTextEntry2, doing all the world->screen
+// projection, distance fade, occlusion (don't show through track geometry) and splitscreen-pass
+// handling. We want the same labels in multiplayer but showing the human player's name instead of
+// the number -- so rather than re-implement that projection (and its ~13 tuning constants), we
+// reuse it verbatim and only swap the drawn string (and scale/nudge its draw position).
+//
+// swrPlayerHUD_RenderDistanceText_delta wraps the renderer: in single-player it just calls the
+// original (numbers over AI untouched); in multiplayer it builds a per-slot name label, sets a
+// redirect flag, runs the original, then clears the flag. While the flag is set,
+// swrText_CreateTextEntry2_delta (which the original calls for each visible, non-occluded racer)
+// maps the draw back to its racer slot by the screen position the renderer just stored, swaps in
+// that slot's name and adjusts the draw x/y.
+//
+// Slot mapping: the sprite slot is the racer entity's obj.id; in multiplayer the swrScores[] index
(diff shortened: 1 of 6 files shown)


dinput_hook/game_deltas/swrPlayerHUD_delta.cpp
```diff
@@ -0,0 +1,185 @@
...
+static int g_labelPosPct = 200;
+static int g_labelOffsetX = 0;
+static int g_labelOffsetY = 0;
+
+static int read_ini_int(const wchar_t *key, int fallback, const wchar_t *ini) {
+    wchar_t buf[32];
+    if (GetPrivateProfileStringW(L"settings", key, L"", buf, (DWORD) std::size(buf), ini) == 0)
+        return fallback; // key absent -> default
+    return _wtoi(buf);    // _wtoi handles a leading '-' (GetPrivateProfileIntW does not)
+}
```
tim-tim707: Why do we need this

dinput_hook/game_deltas/swrPlayerHUD_delta.cpp
```diff
@@ -0,0 +1,185 @@
...
+    loaded = true;
+
+    wchar_t module_path[1024];
+    GetModuleFileNameW(nullptr, module_path, (DWORD) std::size(module_path));
+    const std::wstring ini =
+        (std::filesystem::path(module_path).parent_path() / "SW_RACER_RE.ini").wstring();
+
+    g_labelPosPct = read_ini_int(L"name_label_pos_pct", 200, ini.c_str());
+    g_labelOffsetX = read_ini_int(L"name_label_offset_x", 0, ini.c_str());
+    g_labelOffsetY = read_ini_int(L"name_label_offset_y", 0, ini.c_str());
```
tim-tim707: Why do we need this

dinput_hook/game_deltas/swrPlayerHUD_delta.cpp
```diff
@@ -0,0 +1,185 @@
...
+// the label at half the intended screen coordinate; we undo that by scaling the draw position by
+// 200%.
+//
+// Whether the labels render at all is the "Overhead racer labels" debug-menu toggle
+// (imgui_state.show_pod_names, persisted to SW_RACER_RE.ini as show_pod_names); off hides them in
+// both SP and MP. The remaining placement knobs live in SW_RACER_RE.ini [settings] (edit + relaunch,
+// no rebuild):
+//   name_label_pos_pct  (default 200) percent to scale the draw x/y by (undoes the 0.5 scale)
+//   name_label_offset_x (default 0)   pixels to nudge right (negative = left), after the scale
+//   name_label_offset_y (default 0)   pixels to nudge down (negative = up), after the scale
```
tim-tim707: lets make it constant for now

louriccia:
Addressed all three — dropped the INI tuning plumbing (`read_ini_int` / `load_label_settings_once` + the mutable `g_labelPos*` globals) and baked placement as constants (`HUD_NAME_POS_PCT`=200, `OFFSET_X`/`Y`=0). They were just bring-up knobs; values are dialed in now. cf122ff

PR #101, opened 2026-06-15
swrObjJdge: reimplement F4 (race-judge sub-event dispatcher) + type the struct

louriccia:
Reimplements `swrObjJdge_F4`, the race judge's sub-event dispatcher, and types the bits of the struct it proves.

**`F4`** switches over 13 four-char sub-events: `Begn` (latch the race config -- players / planet / laps / lap records / spline + cam ids), `Load` (reset all per-race state), `Join`/`Mstr` (MP host claim), `Paws` (abort/restart from the pause menu), `Slep`/`Wake` (toggle the entity skip flag), `JAsn` (resolve a score's finish time to its pod object + forward), `RSet`. All 13 event-tag literals static-assert-match the original hex; compiles with the project flags (both `-Werror` gates).

**`types.h`** (layout-preserving, sizeof 0x1e8): `unk1a0` `int`->`float` (stores 1.0 at `Load`), `unk1c4` `char[4]`->`int` (race-config word at `Begn`); documented the `flag` bitfield + `unk64_mat`/`unk12c`/`unk1cc_ms`.

**Globals + helpers `F4` touches:**
- Carved + named `swrSound_SelectPlanetIntroMusic` (0x427ad0 -- arms the per-planet intro music) and `swrScene_SetObjectsLoaded` (0x4804a0); plus `swrEvent_Broadcast` (0x450be0).
- Race-state globals: `swrRace_demoMode`, `swrObjJdge_localRacerId` / `demoHudCycleIndex` / `demoHudCycled` / `creditsScrollState` / `postRaceHudState`, `swrControl_uiInputActive`, `swrPlayerHUD_lightStreakParam`.

**Confidence / flags:**
- High: `F4` control flow + dispatch (hex-verified), the calls to already-named functions, the two field-type fixes, `swrSound_SelectPlanetIntroMusic`.
- Medium (behavioral proposals -- happy to match your DB names): the race-state globals, `swrControl_uiInputActive`, `swrPlayerHUD_lightStreakParam`.
- `swrScene_` is a **provisional namespace** (the first `swrScene_` symbols in src -- open to `swrMain_`/`swrWorld_` if you prefer). `swrEvent_Broadcast`/`swrEvent_BroadcastFiltered` are **vestigial** (build the args then no-op via `swr_noop2` in the release build).

PR #161, opened 2026-06-22
swrObjJdge: reimplement the in-race HUD + race-lifecycle state machine

louriccia:
Reimplements the swrObjJdge race-judge HUD + race-lifecycle state machine (the F4 sibling from #101) -- 13 functions, all reverse-hooked:

- **State machine:** `F0` (per-frame race lifecycle: countdown -> race -> finish -> post-race camera/results -> teardown), `F3` (per-frame HUD render + music fade)
- **Standings/logic:** `UpdateStandings`, `GetRacerProgress`, `GetRacerRankValue`, `CycleHudMode`
- **Render helpers:** `DrawRaceHUD`, `DrawHudBar`, `DrawSplitDivider`, `HideEngineUI`, `UpdateMinimap`, `UpdatePlayerHUD`, `UpdateCountdownLights`

**Verification:** full `libdinput_hook.dll` link (WinLibs GCC 13.2); `-O0` call-counts checked against the Ghidra disasm; all 6 `F0` event tags `_Static_assert` their hex. Broken float ramps the decompiler lost as `unaff_ESI/EDI` were recovered from the x87 disasm.

**Callees:** the reimpl calls a number of functions that were header-only (named, no body). I reimplemented the ones with a clean decompile and no hard dependency cascade -- **6 real bodies**:

- `GetPauseMenuScrollInOut`, `swrMultiplayer_SetNetworkTick`, `swrText_ShowTimedMessage`, `swrObjJdge_IsRacerRacing`, `swrObjJdge_StartPostRaceSequence`, `enablePause`

The remaining **7 stay `HANG("TODO")` reverse-hook stubs** (same pattern as #101), because a faithful body needs work that's out of scope here:
- broken-FP decompiles needing an at-machine x87 pass: `swrObjJdge_ScrollCredits`, `swrSprite_AddDirtyRect`
- float-heavy with many unnamed consts: `swrRace_InitFireEffects`, `swrSpline_CollectNearbyPoints`
- the MP player-status cluster (`InitPlayerStatus`/`SetPlayerStatusBit`/`PollPlayerStatus`), whose `PollPlayerStatus` bottoms out in DirectPlay `PumpPackets` -- belongs in its own focused swrMultiplayer PR
- `swrObjJdge_TeardownRace`: clean logic but reaches ~6 still-opaque globals + two unnamed `FUN_` setters; deferred rather than ship guessed names

**Globals:** new in `data_symbols.syms` -- `swrMain_raceActiveForUi`, `swrObjJdge_unkCa0c` and `pauseEnabledFlag` (both write-only/opaque, named for context -- rename welcome), the timed-message state (`swrText_timedMessage*`), and the planet1/track3 not-racing region bounds (`swrObjJdge_notRacingZone*`).


src/Swr/swrObj.c
```diff
@@ -267,16 +294,475 @@ int GetLocalPlayerNumberFromScore(swrScore*)
...
+    return progress;
+}
+
+// Sort key for standings: finished racers rank by (10000 - total time); others by progress.
+// 0x0045d480
+float swrObjJdge_GetRacerRankValue(swrScore* score)
+{
+    if ((score->flag & 2) == 0)
+        return swrObjJdge_GetRacerProgress(score);
+    return 10000.0f - score->results_P1_total_time;
```
tim-tim707: isn't 10000.0f the constant for 50 minutes max time ?

tim-tim707 (commented):
Looks good.
Would this still work with timer > 50 minutes ?

src/Swr/swrObj.c
louriccia: Good catch -- that 10000.0f is `_DAT_004ad230` in the original (this fn is its only xref). Named it `swrObjJdge_maxRaceTime` and it now references that instead of the literal.

It's just the rank baseline for finished racers: key = `maxRaceTime - total_time`, so the fastest finisher gets the highest value and finished racers stay sorted above the progress-based keys of the unfinished ones. Byte-faithful to vanilla -- if `total_time` ever climbed past the ceiling a finished racer would sort below an unfinished one, but that's the original's behavior and 10000 is well beyond any real race time, so it never triggers in practice. Happy to rename if you've got a better term for it.

louriccia: Ah, re #97 -- you're right to connect them, and it's a real gap. #97 raised the F2 time clamp at `0x4ad264` from 3000.0f (the 50:00 cap) to 86400.0f (24h), but left this rank baseline (`_DAT_004ad230` = 10000.0) untouched. In vanilla `total_time` was pinned to 3000s, so `10000 - total_time` was always >= 7000 and finished racers always outranked the unfinished (progress) ones. With #97's 24h cap a race past ~2h46m (10000s) makes a finished racer's key go negative -> it'd sort below a still-racing one in the live standings. A long 125-lap race can realistically get there.

This reimpl stays byte-faithful to vanilla (keeps 10000.0), so the fix belongs as a #97 follow-up in the delta layer, not here. Simplest is to also bump the baseline at `0x4ad230` to match the 24h cap (the same one-line patch you already did for the 3000s clamp at `0x4ad264`) so `baseline - total_time` stays positive; or make the standings sort finished-first (finished flag before the time key) for a length-independent fix. I'll send a follow-up against the #97 delta -- leaning toward the baseline bump since it mirrors what's already there.

louriccia: Sent the follow-up: #164 hooks swrObjJdge_UpdateStandings to re-assign positions finished-first (finished racers always above still-racing ones, ordered by total_time at full precision), so the standings stay correct past the old 50:00 ceiling. This PR keeps the faithful 10000.0 baseline as-is.

PR #168, opened 2026-06-23
swrText: reimplement swrText_GetStringWidth and swrText_GetStringHeight

louriccia:
Reverse-hook reimpls of the two glyph-metric helpers (`0x0042de30` / `0x0042df70`), verified faithful against the disassembly.

Brings along the types + globals they need:
- `swrFont` (0x68) + `swrTextGlyph` (0x10) structs in `types.h` — layout from `swrText_InitFonts` (5 font structs, stride 0x68) plus the field offsets already documented in `swrText.h` (advance @0x02, height @0x0e, firstChar @0x5a, lastChar @0x5b, glyph tables @0x5c / @0x60).
- Retypes both prototypes `void*` -> `swrFont*`.
- Names the two extended-char (code > 0x96) compose tables: `swrText_extCharComposePairs` (0x4bfa10) and `swrText_extCharComposeIndex` (0x4bfa59). The decompiler's `0x4bf9c2` base is back-computed and lands inside the last font struct; the real index data starts at `0x4bf9c2 + 0x97 = 0x4bfa59`, indexed `[code - 0x97]`, which folds back to `0x4bf9c2` in codegen.

Table byte values weren't directly readable in my setup, so the table sizes are inferred from the layout gap (noted in `data_symbols.syms`); the reimpl only depends on their addresses + indexing, which the disasm confirms.

Builds + links clean.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -175,6 +175,14 @@ lensFlareSpriteIds 0x004B94E0 int[16]
 
 currentSpriteColor 0x004BFA0C uint8_t[4]
 
+# extended-character (code > 0x96) compose tables read by swrText_GetStringWidth/Height/DrawString.
+# A code's row index = swrText_extCharComposeIndex[code - 0x97] (0xff means no entry); that row in
+# swrText_extCharComposePairs is {extGlyphSlot, baseChar}. baseChar 0xff => use the font's extended
+# glyph table at extGlyphSlot, else fall through to the normal glyph for baseChar. Sizes are bounded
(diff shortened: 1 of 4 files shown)


tim-tim707 (commented):
Ready when rebased

louriccia:
Rebased onto current master. The only conflict was in swrText.c where master had added swrText_ShowTimedMessage at the same spot - kept both. Builds + links clean, both reimpls register as reverse hooks.

PR #170, opened 2026-06-23
fix: keep // 0xADDR hook annotations directly above their signature

louriccia:
`GenerateHooks.py` registers a hook by matching the `// 0xADDR` comment and then reading the **very next line** as the signature. If a description comment (or a blank line) sits between the annotation and the signature, the function is silently dropped from the hook table — it still compiles, but it's no longer registered/tracked as a (reverse) hook.

Found **10 cases across 6 files** in that state. This moves each annotation back onto the line directly above its signature (description stays just above it), and removes one **stray duplicate** `// 0x00404a20` in `sithMulti.c`: that address already annotates `sithMulti_CreatePlayerFromConfig` a few lines above, and the stray copy sat above a blank line annotating nothing (`sithMulti_RunCallback` just below it stays correctly hooked via its own `0x0041b8f0`).

Comments only — the disassembly is unchanged. After the fix all affected functions register correctly (verified by regenerating `hook_generated.c`).

Affected: `stdPlatform_DebugLog`, `stdColor_ColorConvertOneRow`, `rdModel3_DrawMesh`, `swrModel_PointInTriangle`, `IntersectRayPlane`, `swrModel_CollideRayTriangle`, `swrModel_MeshCollisionFaceCallbackIndexed`, `swrModel_MeshCollisionFaceCallback`, `swrText_UnescapeString` (+ the `sithMulti.c` duplicate).

Also adds `scripts/check_hook_annotations.py` to catch regressions — mirrors the existing dup-check scripts and exits non-zero on any misplaced annotation, so it could be wired into `integrity-check`.

Builds + links clean.

### DIFF
--- a/scripts/check_hook_annotations.py
+++ b/scripts/check_hook_annotations.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+"""
+Check that every "// 0xADDR" reverse-hook (and "// 0xADDR HOOK" forward-hook) comment
+sits DIRECTLY above the function signature it annotates.
+
+GenerateHooks.py recognizes a hook annotation, then expects the VERY NEXT line to be the
+function signature. If a description comment (or a blank line) sits between the "// 0xADDR"
+line and the signature, GenerateHooks silently drops the function from the hook table -- the
+reimpl still compiles, but it is no longer registered / tracked as a hook.
+
+This catches that before it ships. Run from the repo root:
+    python scripts/check_hook_annotations.py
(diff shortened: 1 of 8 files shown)


scripts/check_hook_annotations.py
tim-tim707, line 1: Add this in the CI please

louriccia: Done - added it to the validation step in integrity_check.yml, right after the two dup-check scripts. Passes locally.

PR #177, opened 2026-06-23
swrMultiplayer: reimplement the player-name getters

louriccia:
Two leaves from the multiplayer roster.

- `swrMultiplayer_GetPlayerName` (0x41bcc0) - returns a pointer to a player's wide display name. Reuses the existing `unicode_unk` global (the roster name field), indexed by the 0xb0 per-player stride. Zero new globals.
- `swrMultiplayer_GetPlayerNameAscii` (0x41bce0) - converts that name to ASCII into a scratch buffer, now named `swrMultiplayer_asciiNameBuffer` (0x4ead88, char[32] matching the 0x20 convert limit).

Both disasm-verified (the `i * 0x58` wchar offset is the 0xb0 byte stride). Builds + links.

While mapping this I confirmed the SWR per-player slot layout (base 0xe9f3c0, stride 0xb0: name1 @ +0x04, name2 @ +0x44, flags @ +0x84 with bit0 = active, dpid @ +0x88). The existing `sithPlayer_g_aPlayers` symbol sits at +0x88 of that slot with the JK `SithPlayer` typedef, which is why its layout "doesn't fit" - happy to follow up with a proper SWR player struct if you'd like.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -1254,7 +1254,7 @@ swrScene_animations 0x00e9edc0 swrModel_Animation*[300] # related to models
 stdPlatform_hostServices 0x00e9f280 HostServices
 
 rootPathName 0x00e9f300 char[80]
-unicode_unk 0x00e9f3c4 wchar_t[32]
+swrMultiplayer_playerNames 0x00e9f3c4 wchar_t[32] # base of the per-player unicode name table (see swrMultiplayer_GetPlayerName)
 
 sithPlayer_g_aPlayers 0x00e9f448 SithPlayer[1] # how many players ?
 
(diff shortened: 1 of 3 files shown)


tim-tim707:
If `unicode_unk` is understood it should be renamed.

Happy with a new player struct if it matches the game

louriccia:
Renamed `unicode_unk` -> `swrMultiplayer_playerNames` and gave the `0x58` per-player stride a name too. (241bedb)

I held off on a player struct for now: that base is the name field of a per-player record that physically interleaves with `sithPlayer_g_aPlayers`, and only the name itself is nailed down (<=15 wchar + NUL at index 0xf, `0x58`-wchar stride to the next entry). The rest of the record isn't established, so a struct would be mostly guesswork. Happy to fold the names into a proper player struct as a follow-up once the SithPlayer layout is mapped.

PR #180, opened 2026-06-23
swrSprite: reimplement swrSprite_AssignTextureToId

louriccia:
One of the `HANG("TODO, easy")` stubs.

`swrSprite_AssignTextureToId` (0x416fd0) finds the first free `swrSpriteTexItems` slot (`texture == NULL`) and records the texture, its sprite id, and the TGA-source flag in the parallel `swrSpriteTexIsTGA` array. Uses the same array walk (and `0x4d8110` end bound) as the existing `swrSprite_FreeSprites` in this file.

Disasm-verified, no new globals. Builds + links.

### DIFF
--- a/src/Swr/swrSprite.c
+++ b/src/Swr/swrSprite.c
@@ -124,7 +124,7 @@ void swrSprite_AssignTextureToId(swrSpriteTexture* spriteTex, int id, int from_t
         }
         texItems = texItems + 1;
         texIsTGA = texIsTGA + 1;
-    } while ((int)texItems < 0x4d8110); // swrSpriteTexItems Array End
+    } while (texItems < swrSpriteTexItems + (sizeof(swrSpriteTexItems) / sizeof(swrSpriteTexItems[0])));
 }
 
 // 0x00417010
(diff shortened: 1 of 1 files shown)


src/Swr/swrSprite.c
```diff
@@ -110,7 +110,21 @@ void swrSprite_ClearSprites(swrUI_unk* swrui_unk)
...
+    do {
+        if (texItems->texture == NULL) {
+            texItems->texture = spriteTex;
+            texItems->id = id;
+            *texIsTGA = from_tga;
+            return;
+        }
+        texItems = texItems + 1;
+        texIsTGA = texIsTGA + 1;
+    } while ((int)texItems < 0x4d8110); // swrSpriteTexItems Array End
```
tim-tim707: Can we use a compile time value instead of a magic ? Like swrSpriteTexItems + itemSize * sizeof(item) ?

louriccia: Good call. Swapped both loops (this one + the same magic two functions down in `swrSprite_FreeSprites`) to `swrSpriteTexItems + sizeof(swrSpriteTexItems) / sizeof(swrSpriteTexItems[0])` — the generated `swrSpriteTexItem[149]` decl folds that to a compile-time bound. Builds + links clean. (939d0fb)

PR #182, opened 2026-06-24
swrText: SDF text renderer (crisp TTF typography, opt-in)

louriccia:
Opt-in crisp in-race text (default **OFF** = byte-faithful vanilla). Reverse-hooks `swrText_RenderString` and lays strings out with a bundled TTF's own metrics + kerning from a packed SDF glyph atlas (DejaVu for body, Anton for the display face), drawn through an SDF shader. Toggle: **"Crisp text (SDF)"** in graphics settings (persisted to the ini).

**Highlights**
- `swrText_InitFonts` reimplemented (advances the decomp); `swrFont` / `swrTextGlyph` added to `types.h` with offsets verified byte-for-byte via read_memory.
- SDF atlas rasterizes on a **worker thread** -- no launch stall; text renders with the built-in bitmap fonts until the atlas is ready (~1s), then upgrades. GL upload stays on the main thread.
- Overhead racer labels (SP position numbers + MP names): small + centered, a resolution-independent lift that scales with the pod's apparent size, and an occlusion debounce (the game's 16-bit depth read flickers at distance).
- Composes with the resolution-independent UI (reads the same design-scale recips).

**Note:** this branch also carries the per-axis Hor+ overlay-reprojection fix submitted standalone as #229 (the labels need it to sit on their pods). Once #229 lands this rebases and the change dedupes.

Built + links clean (WinLibs GCC 13.2, RENDERER_REPLACEMENT). `FONT_RENDERING_ROADMAP.md` has the RE-grounded design + status.


louriccia:
Pushed a follow-up on top of the renderer: an **SDF Fonts** debug panel (per slot: font file, weight, italic/shear, scale, offset, line height, letter spacing, shadow) plus **shareable font profiles** — one `.ini` file each under `assets/fonts/profiles/`, load/save/share from the panel. Default config renders identically to before; it's all opt-in behind the existing toggle. Shipped two profiles: Classic (built-in fonts) and Aspyr (adds Orbitron-Black, OFL). Engine went from two shared faces to a small atlas cache keyed by (path, shear); live tunables update per frame, font/italic changes async-rebuild with a no-flicker swap.

It's a chunky add — happy to split the panel+profiles into its own PR if you'd rather keep this one focused on the core renderer. Just say the word.

louriccia:
Another pass:
- Consolidated all text/font options into one **Font** panel section, and each slot now shows a "used for ..." line (from the ~f->slot mapping) so it's clear which draws share a font.
- **OTF/CFF**: warn + amber hint when an .otf is picked (stb_truetype's CFF support is partial), and fall back to `assets/fonts/<name>` when a path doesn't resolve so shared profiles stay portable.
- Shipped profiles: **LucasArts** (built-in), **Aspyr** (Orbitron, OFL), **Fuse** (Moki + Aurebesh, bundled with permission).
- Dropped the roadmap doc from the PR.

Two known items I'm leaving as separate follow-ups so this stays scoped: (1) the text-input **caret** is positioned by the bitmap-font substring width (`swrUI_GetSubstringWidth`), which the SDF metrics no longer match — same family as the wider UI-metrics work; (2) MP chat rejecting punctuation is a **vanilla** filter in `swrMultiplayer_HandleChatKey`, not the fonts.

PR #184, opened 2026-06-25
Debug overlay: panel registry + Quick Race / Settings / Cheats

louriccia:
image

Reworks the F5 debug overlay from the single flat `opengl_render_imgui()` into a small
panel registry, then builds a bunch of player-facing tools on top. All in the `dinput_hook`
overlay layer -- no `src/` reimpl changes.

**Shell** (`debug_ui.{h,cpp}`): each subsystem registers a `DebugPanel`; the shell draws one
"SWE1R Debug" window of collapsing-header sections grouped by category, with a section filter,
expand/collapse-all, an FPS sparkline, and a Developer-mode toggle (dev sections hidden in
release). This kills the old "every feature splices the same two spots" merge-conflict problem.

**Sections**
- Render: graphics settings, HD models
- Quick Race: pod + track dropdowns, AI racers, laps, mirror, AI speed, winnings, and a
  "Restart race" button that applies them via the game's own reload (`swrObjJdge_Clear(jdge,'RStr')`,
  reached through `swrEvent_GetItem`) -- same approach annodue uses, no live entity surgery
- Settings: Audio (master via `swrSound_SetOutputGain`, SFX/music, 3D/doppler/voices, hi-res),
  Video (the native `swrConfig_VIDEO_*` flags), Controls (joystick enable + deadzone)
- Cheats: god / no-overheat / no-OOB / fly / fast, unlock-all, +truguts (applied every frame
  from `imgui_Update`)
- Inspect (dev): Scene, Textures, Pod Transforms, Pod Readout telemetry
- Tools: theme/opacity/demo/metrics; the hook log is back as its own toggleable window

HD-replacement and the F5 visibility flag now persist to `SW_RACER_RE.ini`.

Built + linked clean; playtested. Happy to split it up if you'd rather review the shell and the
features separately.

### DIFF
--- a/dinput_hook/debug_ui.cpp
+++ b/dinput_hook/debug_ui.cpp
@@ -1,0 +1,141 @@
+#include "debug_ui.h"
+#include "imgui_utils.h"
+
+#include <vector>
+#include <string>
+#include <cstring>
+#include <cstdio>
+
+#include <windows.h>
+#include <imgui.h>
+
+// show_imgui (the F5 overlay toggle) and settings_ini_path() come from imgui_utils.h.
+
+#if !defined(NDEBUG)
+bool debug_ui_show_dev_panels = true;
+#else
+bool debug_ui_show_dev_panels = false;
+#endif
+
+static std::vector<DebugPanel *> g_panels;
+
+void debug_ui_register(DebugPanel *panel) {
+    g_panels.push_back(panel);
+}
+
+// Panel names and ini keys are ASCII literals we control, so a byte-wise widen
+// is enough to feed the wide GetPrivateProfile* API (matches the ini path type).
+static std::wstring widen(const char *s) {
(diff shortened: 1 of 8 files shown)


tim-tim707 (changes requested):
Expand All / Collapse all have some overflow issue on the right

Developer mode should be off by default
FPS show graph of by default

louriccia:
Addressed all three:

- **Developer mode** now defaults off in every build (it was defaulting on in debug builds, which is likely what you hit).
- **FPS overlay graph** defaults off now -- it's opt-in via the FPS panel's "Show graph" toggle.
- **Expand all / Collapse all** -- the filter row now reserves the real label + button widths instead of a fixed amount, so the buttons stay inside the window instead of clipping on the right.

Also merged latest master and built clean.

PR #186, opened 2026-06-25
Windowed mode: re-render on resize + aspect-correct FOV + FOV slider

louriccia:
Fixes windowed mode. Resizing the window now re-renders the game at the new resolution instead of stretching a fixed-size frame, and the camera FOV is aspect-correct for non-4:3 with a user slider.

- A GLFW framebuffer-size callback keeps `swrDisplay_screenWidth/Height` tracking the live window size. The GL viewport + MSAA FBO already follow the framebuffer each frame; these globals were the only thing frozen at startup, so the 3D projection, the 2D ortho, and the cursor all rendered for the startup size and got stretched into the resized window.
- The projection now holds the 4:3 **vertical** FOV constant across aspect ratios (Hor+) instead of the horizontal one, so widescreen reveals more horizontally rather than cropping top/bottom. 4:3 is byte-identical to before.
- New persisted **FOV scale** slider (0.5-2.0, default 1.0).

Independent of #185 (this is a 3D-camera / window concern). Builds + links clean; playtested in windowed mode at various sizes and with the FOV slider.

### DIFF
--- a/dinput_hook/game_deltas/Window_delta.c
+++ b/dinput_hook/game_deltas/Window_delta.c
@@ -388,6 +388,19 @@ void GLAPIENTRY Window_glDebugMessageCallback(GLenum source, GLenum type, GLuint
 }
 
 // 0x0049cd40
+// On a live window resize, keep swrDisplay_screenWidth/Height (== screen_width/height) tracking the
+// framebuffer. The GL viewport + MSAA FBO already follow the framebuffer each frame; these globals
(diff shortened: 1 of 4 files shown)


louriccia:
Merged current `upstream/master` in to bring this up to date (and kick the integrity-check CI, which hadn't run on the original push). Conflicts were just the new `show_pod_names` setting landing next to the FOV-scale load/slider in `imgui_utils.cpp` — kept both. Builds + links clean.

PR #202, opened 2026-06-26
fix: render rain/snow at modern resolutions

louriccia:
Rain/snow don't render on modern high-res displays. Two reversible EXE byte-patches in a new `swrWeather_delta`, applied at renderer init:

**Layer 2 - no weather at screen_width >= 1000.** `swrViewport_ProjectToScreen` seeds its off-screen sentinel at `-1000.0f` and the particle despawn test is `screenX < -screen_width`, so at `screen_width >= 1000` the despawn never fires (`-1000.0 < -1000.0` is false), the pool stops recycling, and weather vanishes after one cycle. Patch the two sentinel immediates to `-INF` so despawn fires at any resolution. (The old dgVoodoo theory is wrong - DDrawCompat reproduces the same threshold, so it's the EXE.)

**Layer 3-A - snow/rain vanish the moment the pod moves.** The per-particle render mode is chosen by an absolute 3px screen-movement threshold; at high res any camera motion exceeds it, flipping every particle to the streak path (sprite flag `0x4000`). That path turns out to be a stub: instance flag `0x4000` -> draw flag `0x400000` (`swrSprite_Draw2`) -> `swrSprite_Draw` routes it to `swr_noop2` (a single `RET`), so a streaking sprite emits no geometry at any resolution - the PC port never implemented the N64/Dreamcast motion-blur streaks. NOP the two `JGE` branches that select the streak path so the working point-sprite path always runs; moving weather then renders as points (PC's only-ever behaviour).

Both patches byte-verify the site before writing and no-op (logged) if the signature doesn't match. Also corrected the Layer 3 notes in `swrWeather.h` (they guessed a precision bug in the streak math; it's actually the `swr_noop2` stub).

Playtested at 1920x1080: snow and rain now render both parked and moving. Real motion-blur streaks would be a separate feature (reimplementing the cut path), and rain is still a touch sparse (Layer 3-B, noted in the header) - both good follow-ups rather than blockers here.

### DIFF
--- a/dinput_hook/game_deltas/swrWeather_delta.cpp
+++ b/dinput_hook/game_deltas/swrWeather_delta.cpp
@@ -1,0 +1,101 @@
+#include "swrWeather_delta.h"
+
+#include <windows.h>
+#include <cstdint>
+#include <cstdio>
+#include <cstring>
+
+extern "C" {
+extern FILE *hook_log;
+}
+
+// LAYER 2: lift the high-resolution weather-despawn deadlock. See swrWeather_delta.h and the
+// KNOWN ISSUES block in src/Swr/swrWeather.h for why this exact byte edit fixes it.
+//
+// The two patched dwords are the float immediates inside these instructions:
+//   0x0042B868  MOV dword ptr [ESI], 0xC47A0000   ; *outScreenX = -1000.0f   (imm @ 0x0042B86A)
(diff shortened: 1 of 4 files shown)


louriccia:
Superseded by #203, which folds this high-res fix in and reimplements the cut motion-blur streaks (smooth round particles, depth occlusion, graceful SNW/NSNW transitions). Closing in favour of the combined PR.

PR #204, opened 2026-06-28
feat(multiplayer): per-player pod upgrades

louriccia:
Vanilla multiplayer races everyone on raw base stats -- `swrObjHang_BuildRosterMultiplayer` copies each pod's base `swrRacer_PodHandlingData` and never calls `swrRace_ApplyUpgradesToStats` (unlike the single-player builder), and MP has no pilot-profile step to pull upgrades from.

This adds an opt-in **"Multiplayer: allow pod upgrades"** toggle plus per-category level sliders (0..5: traction / turning / accel / top speed / air brake / cooling / repair). When it's on, the local player's chosen upgrades are layered onto their own `'Locl'` score entry after the roster is built, at full part condition, through the game's own `swrRace_ApplyUpgradesToStats` so the resulting stats stay valid and clamped. Remote pods are transform-replayed, so each machine only upgrades its own pod -- everyone races with their own picks.

Per-player by design (no host sync, kept simple on purpose); default off; single-player untouched. `swrObjHang_BuildRosterMultiplayer` isn't reimplemented, so it's hooked by address. Playtested solo (maxed top speed hits the 650 cap).

### DIFF
--- a/dinput_hook/game_deltas/swrMultiplayer_delta.cpp
+++ b/dinput_hook/game_deltas/swrMultiplayer_delta.cpp
@@ -8,6 +8,8 @@ extern "C" {
 #include <Dss/sithMulti.h>
 #include <Win95/stdComm.h>
 #include <Swr/swrUI.h>
+#include <Swr/swrObj.h>
+#include <Swr/swrRace.h>
 #include <Swr/swrMultiplayer.h>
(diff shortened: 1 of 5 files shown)


tim-tim707:
still waiting on rebase here

Issue #153, opened 2026-06-21
Better modding API


tim-tim707:
In order to toggle mods, we need unhook functions and unpatch functions.
Add an api function to write to a page with VirtualProtect that would do bookeeping to save the bytes written to in order to undo later.

WriteMemory(address, bufferToWrite, bufferLen) (would internally save the previous bytes so that we can undo the writeMemory later)
Also check that we don't write to the same memory twice or hook the same address with two different functions.
On a second phase, we can add unhook_function and unreplace_hook functions.

Also replace the settings with a simple key-value parser that support writting comments that would replace the GetModuleFileNameW GetPrivateProfileIntW WritePrivateProfileStringW api.
These readSettings / writeSettings would be hooked in the game and listen for changes in the config file to auto-refresh the settings.
This would enable setting modifications without imgui or a new renderer.

louriccia:
Spent some time on this and looked at how annodue handles it, plus the patterns in Harmony (.NET), Forge/Fabric Mixin, and WoW. Proposed shape:

**1. Reversible patches.** `WriteMemory(addr, buf, len)` backed by a small undo journal that captures the true original bytes once and refcounts overlapping ranges, so any write can be reverted. `unhook_function` / `unreplace_hook` ride on top - `DetourDetach` for the jmp hooks, journal-revert for the raw byte patches. I'd port `set_ai_full_lod` onto it first since it's already a `{addr, len, original[]}` toggle.

**2. Mod registry.** Each feature = `{name, enable(id), optional on_disable()}`. `disable` just replays that owner's journal in reverse, so nobody hand-writes a mirror un-patch (the thing annodue/Blender both make you do by hand). The existing deltas (ai_full_lod, 100-lap, splitscreen, HD pods, overhead names) become registered mods; the always-on reverse-hooks stay "core".

**3. Config.** Comment-aware `key = value` parser replacing the `GetPrivateProfileIntW` / `WritePrivateProfileStringW` calls, plus a file-watch that diffs and applies mod toggles on a frame boundary. That's the "settings without imgui" piece.

Two forks I'd want your call on before building, since they change the API shape:

[...]

PR #209, opened 2026-06-29
feat(modding): reversible memory-patch journal (WriteMemory/UndoOwner)

louriccia:
First step toward the modding API (#153): a reversible, owner-tagged memory-patch journal so mods can mutate game memory without silently corrupting each other or leaking patches on disable.

**`dinput_hook/patch.{h,cpp}`:**
- `WriteMemory(owner, addr, src, len)` - captures the true original bytes once, then patches. Refuses (and logs) a write that overlaps a *different* owner's live patch - that's the corruption guard.
- `PatchPointer(owner, addr, value)` - journaled dword poke (the revertible form of the old `patchMemoryAccess`).
- `UndoOwner(owner)` - restores an owner's captures newest-first.

**Routed the existing raw `.text` patches through it - no behavior change:**
- `set_ai_full_lod` -> `WriteMemory`/`UndoOwner` (drops the hard-coded stock bytes; the journal captures live originals).
- `patchMemoryAccess` -> `PatchPointer` (no callers today; future-proofed).
- the `Window_Main` boot reroute -> two `WriteMemory` calls.

Playtested: normal boot + toggling `ai_full_lod` on/off/on mid-session both behave exactly as before.

Naming (`WriteMemory`/`PatchPointer`/`UndoOwner`) follows the #153 discussion - happy to namespace/prefix if you'd prefer. The other raw-patch features (100-lap de-index, race-time cap, swrModel texture buffer) can migrate onto this in a follow-up; they already hand-roll the verify-and-skip logic this centralizes.

### DIFF
--- a/dinput_hook/patch.cpp
+++ b/dinput_hook/patch.cpp
@@ -14,7 +14,7 @@ extern "C" FILE *hook_log;
 
 namespace {
     struct JournalEntry {
-        PatchOwner owner;             // matched by string value, not pointer
+        PatchOwner owner;             // compared by value via same_owner (pointer-equal fast path)
         uintptr_t addr;
         std::vector<uint8_t> original;// bytes snapshotted immediately before this write
     };
--- a/dinput_hook/patch.cpp
+++ b/dinput_hook/patch.cpp
@@ -1,0 +1,110 @@
+//
+// Reversible memory-patch journal (modding API, issue #153). See patch.h.
+//
+#include "patch.h"
+
+#include <algorithm>
+#include <cstdio>
+#include <cstring>
+#include <vector>
+
+#include <windows.h>
+
+extern "C" FILE *hook_log;
+
+namespace {
+    struct JournalEntry {
+        PatchOwner owner;             // matched by string value, not pointer
+        uintptr_t addr;
+        std::vector<uint8_t> original;// pristine bytes captured before this range was first patched
+    };
+
+    std::vector<JournalEntry> g_journal;
+
+    bool same_owner(PatchOwner a, PatchOwner b) {
+        if (a == b)
+            return true;
+        return a && b && std::strcmp(a, b) == 0;
+    }
+
+    bool ranges_overlap(uintptr_t a, size_t alen, uintptr_t b, size_t blen) {
+        return a < b + blen && b < a + alen;
+    }
+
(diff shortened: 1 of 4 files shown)


dinput_hook/patch.cpp
```diff
@@ -0,0 +1,110 @@
...
+        VirtualProtect(addr, len, oldProtect, &oldProtect);
+    }
+}
+
+bool WriteMemory(PatchOwner owner, void *addr, const void *src, size_t len) {
+    if (!addr || !src || len == 0)
+        return false;
+
+    const uintptr_t a = (uintptr_t) addr;
+    bool reapply = false;// same owner re-writing a range it already journaled (e.g. a re-toggle)
```
tim-tim707: If we undo a write we should clear the journal entry associated with it.

That way we simplify this reapply logic, and also don't keep any side effect of un-patching on our side.
The api should be a fully inverse function where applying patches and undo-ing them would result in a clean internal state

tim-tim707 (changes requested):

louriccia:
Good call - `UndoOwner` already erased its entries, so I leaned into that and dropped the `reapply` machinery (and the exact-match / partial-overlap sub-cases) entirely. `WriteMemory` now just snapshots the current bytes and pushes; undo unwinds newest-first and drops them. Apply + undo is a clean inverse (journal back to empty) even for repeated/overlapping same-owner writes, and the cross-owner conflict guard is unchanged. -12 lines net. Re-built + retested boot and `ai_full_lod` on/off/on.

dinput_hook/patch.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::vector<uint8_t> original;// bytes snapshotted immediately before this write
+    };
+
+    std::vector<JournalEntry> g_journal;
+
+    bool same_owner(PatchOwner a, PatchOwner b) {
+        if (a == b)
+            return true;
+        return a && b && std::strcmp(a, b) == 0;
+    }
```
tim-tim707, line 28: The comment `matched by string value, not pointer` seem to be a lie since at line 25 we compare pointers.

It should work properly anyway but the comment should be updated

louriccia: Good catch - reworded to `compared by value via same_owner (pointer-equal fast path)`. Pushed in 97aebdf.

PR #215, opened 2026-06-29
Collision / trigger / pod-hitbox debug overlay

louriccia:
Adds an **Inspect → Collision** debug overlay (renderer-replacement layer; sim untouched, toggles default-off). Three independent toggles, shared wireframe/opacity controls:

- **Show collision mesh** — walks the track graph the way the game's own query does (`swrModel_CollideNodeRecursiveRay`, `flags_2` per-planet-track mask) and draws the `collision_vertices` the pod tests against, colored by surface reaction (boost/slip/lava/...).
- **Show triggers + racer markers** — trigger volumes per mesh `behavior->triggers` (plane/capsule) + a marker per racer. Port of [annodue](https://github.com/everalert/annodue)'s trigger viz.
- **Show pod hitbox** — the pod body collision sphere (radius read live per-pod, shrunk by lean angle as in `swrRace_DetectWallScrape`) + the 2D pod-vs-pod radius disc.

Flat-color shader drawn into the scene framebuffer before the blit, with full GL state save/restore so the game UI / ImGui are unaffected. Collision-viewer approach + trigger geometry ported from annodue (tly), MIT. Playtested on Tatooine and others — mesh aligns with terrain, triggers match boost gates, hitbox matches where the pod stops at walls.


tim-tim707:
Can you rebase this please ?

PR #217, opened 2026-06-30
feat(modding): portable round-trip config parser (replace Win32 INI)

louriccia:
Phase 4 of the modding API (#153): a small INI-style config that replaces the Windows `GetPrivateProfile*` / `WritePrivateProfile*` calls with a portable parser that **round-trips the file** - comments, blank lines, key ordering and unrecognised keys all survive a save. It drops the mod layer's dependence on the Win32 profile API (a step toward the eventual decomp / cross-platform) and is the foundation for config-driven mod toggling.

**New `dinput_hook/config.{h,cpp}`:**
- Backs the same `SW_RACER_RE.ini` next to the exe, same `[section]` / `key=value` layout, so existing configs load unchanged.
- `;` and `#` comments; case-insensitive section/key lookups; unicode-safe paths via `_wfopen`.
- `get_int` / `get_float` / `get_string` + `set_int` / `set_float` / `set_bool` / `set_string` + `reload()` / `save()`.

**Migrated every Win32-INI consumer onto it (no behavior change):**
- `imgui_utils.cpp` - `read_settings_ini` / `save_settings_ini` (the `[settings]` block); dropped the duplicated path lambda and `settings_ini_path()`.
- `debug_ui.cpp` - `[debug_ui]` / `[debug_ui_panels]`; dropped the `widen()` helper and `<windows.h>`.
- `swrMultiplayer_delta.cpp` - `mp_async_send` / `mp_packet_cap`; dropped its own path resolution and `<filesystem>`.

No Win32 profile calls remain in the delta layer (net -62 lines in the consumers). The auto-refresh / file-watch half of issue Phase 2 is intentionally deferred - it drives `enable_mod`/`disable_mod`, which land with the registry (roadmap Phase 2).

Built + linked clean (WinLibs i686 GCC 13.2); the three integrity-check scripts pass locally. Playtested: existing ini loads unchanged, settings persist across relaunch, hand-added comments/blank lines survive an in-game save, debug-ui panel state persists, and a missing ini is recreated with defaults.


tim-tim707:
I think this needs an update pass to be in sync with the main branch config ?

louriccia:
Synced. This is now rebased on current master and carries every settings key that landed since: #303's player-camera panel, #296's asset-buffer size read, #273's millisecond-times toggle, #253's reflection-texgen block and #215's collision-viewer settings — all ported onto the portable parser rather than left on `GetPrivateProfile*`. The now-unused `settings_ini_path` shim is gone, so there are no Win32 INI calls left in the delta layer.

Builds and links clean, CI green.

PR #235, opened 2026-07-08
Extensible roster: make Cy Yunga & Jinn Reeso separately selectable

louriccia:
@
Right now the two secret pilots are only reachable via the cheat codes that overwrite Mars Guo / Bullseye in place -- so theyre mutually exclusive with their hosts and the swap leaks onto AI. This makes them real, separately selectable racers (ids 23/24) instead.

The 23-pilot roster lives in three parallel fixed-address, id-indexed tables (`swrRacer_PodData`, `swrRacer_PodHandlingData`, the pod engine/cockpit-xform table) plus the SELECT_VEHICLE list. All four are relocated to larger heap arrays, the stock 23 are byte-copied (ids 0..22 unchanged), and the two pilots are appended, seeded from the existing swap data. Rather than reimplement the ~25 readers, each is repointed by shifting its table-address immediate by the relocation delta -- so the original game logic runs byte-for-byte, only the addresses it reads from move. The site lists come from an exhaustive Ghidra xref scan of the three ranges; the patch is transactional (verify every site, then apply) and journaled through the existing revertible `WriteMemory`. `BuildPartMenuList` is reimplemented to enumerate the extended roster.

They unlock with their hosts (Jinn with Mars Guo, Cy with Bullseye) and can turn up as AI opponents. Its also meant as a foundation for adding more characters later.

Lives entirely in the `dinput_hook` delta layer. Scope is single-player; multiplayer and the 2D pilot portraits (the sprite-band relayout) are follow-ups.

@


tim-tim707:
Looks good but I would move the two `_sites.inc` files into the source directly.

louriccia:
Done -- inlined both site lists into swrRoster_delta.cpp (kPodSites / kSelIndexSites) and dropped the two .inc files. Kept each file's table legend as a leading comment. No functional change; rebuilt + linked clean.

tim-tim707:
Didn't this PR crash in multiplayer ? Just to be sure to guard against it

louriccia:
Good catch — you're right, and it would have. Pushed a guard.

The problem: the racer id crosses the wire raw. `swrMultiplayer_RacerPick` sends it in message 0x33, `swrMultiplayer_ApplyRacerPick` stores it into `multiplayer_racer1_id[]` **with no range check**, and `swrObjHang_BuildRosterMultiplayer` then indexes `swrRacer_PodData` / `PodHandlingData` / the transform table with it directly. This PR only relocates those tables on a machine running this build, so a modded client picking Jinn (23) or Cy (24) walks a stock peer past the end of its 23-entry arrays.

Three guards:
1. **Menu cap** — the SELECT_VEHICLE list stops at the retail 23 while `multiplayer_enabled`, so the appended pilots are single-player-only for now.
2. **Outbound clamp** in `RacerPick` — covers a selection carried in from single-player without reopening the menu.
3. **Inbound clamp** in `ApplyRacerPick` — so a peer on a different build can't walk *us* off the end either.

Making them actually selectable in MP needs the roster negotiated in the protocol (both ends agreeing on the table size before the picks start), which felt like its own PR rather than something to bolt on here.

Unrelated, but while I was in there: `ApplyRacerPick` also writes `swrRace_UnlockDataBase[playerIndex * 0x50]` with an unchecked player index straight off the wire. I left that alone since it's vanilla behaviour and nothing to do with this PR, but it's worth hardening at some point.

PR #239, opened 2026-07-08
reimpl: 17 leaf functions (swrText / swrRender / rd / stdMemory / swrSprite)

louriccia:
A batch of small leaf reimplementations, one commit per subsystem so they're easy to review (or split):

- **swrText** - `CreateTimeEntry` / `CreateTimeEntryPrecise` (race-time formatting)
- **swrRender** - `Direct3d_ConfigFog2`, `rdFace_SetFogEnabled`
- **rd** - `rdModel_SetCurrentMaterial` + the render/geometry/lighting mode setters
- **stdMemory** - the full arena allocator: `daSmallAlloc` + `daAlloc` / `daFree` / `daRealloc` (with the `daArena`/`daBlock` structs, `DABLOCK_*` flags and `DAALLOC_*` constants)
- **swrSprite** - `rdProcEntry_Add2DQuad` / `_Add2DQuad4` / `_Add2DQuad2` + `getCurrentTextPos` / `ResetCurrentMaterial`

`Add2DQuad2` (the text-glyph emitter) needed the swrText bbox/clip/orientation globals named - those are in `data_symbols.syms`.

The time formatters and `Add2DQuad2` were blocked by garbled x87 float->int decompilation; I fixed `__ftol`'s calling convention in the Ghidra DB (input on ST0, return in EDX:EAX), which cleans that up DB-wide.

Full `dinput.dll` build links clean.

### DIFF
--- a/src/Swr/swrText.c
+++ b/src/Swr/swrText.c
@@ -456,28 +456,28 @@ void swrText_CreateEntry2(short x, short y, char r, char g, char b, char a, char
 }
 
 // 0x004505f0
-void swrText_CreateTimeEntryFormat(int x, int y, int unused, int r, int g, int b, int a, int bFormat)
+void swrText_CreateTimeEntryFormat(int x, int y, float time, int r, int g, int b, int a, int bFormat)
 {
     char* screen_text;
 
     if (bFormat != 0)
     {
         screen_text = swrText_Translate("~r~s");
-        swrText_CreateTimeEntryPrecise(x, y, unused, r, g, b, a, screen_text);
+        swrText_CreateTimeEntryPrecise(x, y, time, r, g, b, a, screen_text);
         return;
     }
     screen_text = swrText_Translate("~s");
-    swrText_CreateTimeEntryPrecise(x, y, unused, r, g, b, a, screen_text);
+    swrText_CreateTimeEntryPrecise(x, y, time, r, g, b, a, screen_text);
 }
 
-// Format a race time (the `unused` argument is the time in seconds) as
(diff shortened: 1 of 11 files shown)


src/Swr/swrText.c
```diff
@@ -470,45 +470,52 @@ void swrText_CreateTimeEntryFormat(int x, int y, int unused, int r, int g, int b
...
-    }
+        sprintf(buffer, "%s%d:%.2d.%.2d", screenText, minutes, seconds, centis);
     swrText_CreateTextEntry1(x, y, r, g, b, a, buffer);
 }
 
+// Format a race time (the `unused` argument is the time in seconds) as
+// [minutes:]seconds.milliseconds and enqueue it as a text entry. Millisecond-precision
+// variant of swrText_CreateTimeEntry (%.3d vs %.2d, and a finer rounding bias).
 // 0x00450760
 void swrText_CreateTimeEntryPrecise(int x, int y, int unused, int r, int g, int b, int a, char* screenText)
```
tim-tim707: rename the 'unused' argument to something that makes sens

tim-tim707 (changes requested):

src/Swr/swrText.c
louriccia: Good catch, and it turned out to be more than a naming thing. The arg is the race time in seconds, and the binary loads it with `FLD dword` + `FMUL` (1/60) at 0x00450676, so it's actually a `float`. Typing it `int` + `(float)arg` was reinterpreting the raw float bits as an integer, so I've retyped the 3rd param `float time` and dropped the cast across all three formatters (+ the header). Full dinput.dll still links clean. Pushed.

PR #241, opened 2026-07-09
Widescreen UI: anchor menus + in-race HUD to real screen edges

louriccia:
Builds on the resolution-independent 2D UI toggle to make the menus and in-race HUD fill a wide screen instead of sitting in the centered 4:3 box. All gated behind the existing res-independence toggle (exact vanilla passthrough when off).

**Menus**
- Back/Cancel/Quit + Settings anchor to the real left edge, OK to the real right edge, and now track a live window resize (the shift reconciles each frame in `swrUI_RenderElementSprites`, moving the whole element so chrome + label + hit-test stay together).
- Full-screen backdrops (greyedsplash / podhangar_backdrop / splash) stretch to fill the window instead of pillarboxing; the title logo (episode1/star/wars/racer + trademark text) left-anchors.
- GLFW window floored at 320x240 so a drag-resize can't hand the GL driver a degenerate framebuffer.

**In-race HUD** (each cluster anchored to its corner/edge via a small per-sprite-id + per-text-position table)
- Minimap and speedometer -> right; engine readout -> bottom-left; header bar -> lap left / position right / time center, with the connectors + blue gradient stretched to bridge full width.
- Position markers are mode-aware (the HUD-mode draw is scoped so markers remap by mode): catch-up arrows -> right strip, progress ring -> full-width so the flags travel the whole screen.
- Added a debug-overlay "Cycle HUD mode" button (Race panel) as a Caps Lock alternative, since Caps Lock doesn't emulate over remote desktop.

Playtested across the menus and all single-player HUD modes with a 12-racer field. Splitscreen HUD modes (4-7) left for a follow-up.

---

**Update 2026-09-01 — this is now the default, plus a naming pass**

`Resolution-independent UI` was off-by-default from #185 while the layout work was still landing. Now that the menus and the HUD both anchor properly, widescreen is the better out-of-box behaviour, so this PR flips it on: the vanilla 4:3 stretch is the surprising look on a modern display, not the safe one. The checkbox loses its "(experimental)" tag and reads `Resolution-independent UI (widescreen menus + HUD)`, matching the plain labels the other default-on toggles use.

Scope on that: **fresh installs only.** The ini fallback applies just when the key is absent, and the setting is written on every save, so anyone who has run an earlier build keeps whatever they had. No migration of existing configs — deliberately, since it would rewrite settings people chose. Turning it back off still restores exact vanilla behaviour.

The anchor tables also no longer key off raw sprite ids. `src/types_enums.h` already had a `swrUISprite` enum covering exactly that id space, and its names line up with the roles these clusters turned out to have, so they're used instead of new constants:

| cluster | constants |
|---|---|
| lap / position holders | `dial_lap_pos_rgb_0` / `_1` |
| time holder, top rule | `dial_time_wire_rgb_0` / `_1` |
| lap→time, time→pos connectors | `dial_top_h_wire_rgb_0` / `_1` |
| speedo frame + caps | `dial_speed_mid_blue`, `dial_speed_edge_blue_rgb_0` / `_1` |
| ring rails, meter bars | `dial_trakmid_wire_rgb_0` / `_1`, `dial_traktop_wire_rgb_0` |
| engine damage readout | `engdisp0_rgb_0` .. `engdisp3_rgb_3` |
| per-racer position markers | `pilotSprite_20` .. `pilotSprite_40` |

Two land on `ej_fadel_*` names that describe those slots' front-end role rather than their race-HUD one (`0xd` blue gradient, `0x19` radar backing). The constants are still the right ones — that's the slot reuse the enum's own header warns about ("Some Sprite override each other, between UI and In-game"), and the same reuse `ui_in_race_hud` exists to gate — so they're used with a comment saying so. If you'd rather have race-HUD aliases in `swrUISprite`, that's a `src/` change and I'd keep it to its own PR.

HUD text columns are design-x coordinates rather than sprite ids, so they get their own named set (`kHudTextX*`); that switch had been mixing hex and decimal. `160` became `kSpriteDesignCenterX`, derived from `UI_DESIGN_W`. Also fixes a stale comment in `ui_transform.h` that documented the marker range as `0x2b-0x34` (10 racers) when the code guards `0x2b-0x3e` (20).

Every id substitution was verified value-identical against `types_enums.h`, so the naming commit has no behaviour change. Re-playtested at 1920x1017 across the menus and HUD modes 0/3/4, and confirmed the new default by removing the key from an existing ini and relaunching.


louriccia:
Pushed two commits and updated the description.

- **This is now the default.** It was off-by-default from #185 while the layout work was still landing; with the menus and HUD both anchoring properly, the vanilla 4:3 stretch is the surprising look on a widescreen display rather than the safe one. Fresh installs only — the ini fallback only applies when the key is absent, so anyone who has run an earlier build keeps their current setting, and turning it off is still exact vanilla.
- **The anchor tables no longer use raw sprite ids.** `swrUISprite` in `types_enums.h` already covered that id space and its names matched the roles these clusters turned out to have, so I used it rather than adding constants. Every substitution was checked value-identical, so that commit is behaviour-neutral. Two ids land on `ej_fadel_*` names that describe their front-end role instead of their race-HUD one — flagged in a comment; happy to add race-HUD aliases in a separate `src/` PR if you'd prefer.

Re-playtested at 1920x1017 across the menus and HUD modes 0/3/4, and verified the new default by dropping the key from an existing ini and relaunching. CI green.

No rush on my end, but this one has been sitting since July and the default flip is the last piece — whenever you have a window for it.

PR #243, opened 2026-07-09
reimpl(rdCamera): BuildFOV + BuildClipFrustum

louriccia:
Reimplements the last two `HANG` stubs in `rdCamera.c` — the FOV/clip-frustum setup math. The `__ftol` decompiler fix made these single-conversion FP functions legible again.

**`rdCamera_BuildClipFrustum`** — stub -> full. Plane-normal vector math (`rdVector_Normalize3Acc_2`).

**`rdCamera_BuildFOV`** — the previous body was transcribed from a corrupted (pre-ftol) decompile, and the clean decompile exposes three defects, all verified against the disassembly at `0x48fdc0`:
- ortho branch was writing the wrong clip-frustum fields (shifted by 8 bytes) — it now writes `nearPlane`/`farPlane`/`orthoLeftPlane`/`orthoTopPlane` and zeroes the correct set (this is the "members are shifted" the old TODO flagged).
- perspective `orthoRightPlane` was inverted: `aspect / (h/fovY)` vs. the real `(h/fovY) / aspect`.
- `camera->unk` was truncated via `(int)`, but the binary stores a float (a depth-range normalization scale `1/(fovY/zFar - fovY/zNear)`), so I retyped the field `uint32_t -> float`. It was write-only in `src/`.

Cross-checked the width/height assignment in `fov_y` against OpenJKDF2's `rdCamera_BuildFOV` (`fov_y = xExtent / tan(fov*0.5)`), since the FXCH scheduling makes the decompiler's var ordering ambiguous.

Builds + links clean.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -259,7 +259,7 @@ extern "C"
         void (*fnProject)(rdVector3*, rdVector3*); // 0x4c
         void (*fnProjectLst)(rdVector3*, rdVector3*, unsigned int); // 0x50
         float ambientLight;
-        uint32_t unk;
+        float unk; // 1 / (fov_y/zFar - fov_y/zNear); depth-range normalization scale
         rdVector4 unk2;
         int numLights; // 0x6c
         rdLight* lights[128]; // Jkdf 64, Indy 128. really 128 ?
--- a/src/types.h
+++ b/src/types.h
@@ -259,7 +259,7 @@ extern "C"
         void (*fnProject)(rdVector3*, rdVector3*); // 0x4c
         void (*fnProjectLst)(rdVector3*, rdVector3*, unsigned int); // 0x50
         float ambientLight;
-        float unk; // 1 / (fov_y/zFar - fov_y/zNear); depth-range normalization scale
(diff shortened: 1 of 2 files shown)


src/types.h
```diff
@@ -259,7 +259,7 @@ extern "C"
         void (*fnProject)(rdVector3*, rdVector3*); // 0x4c
         void (*fnProjectLst)(rdVector3*, rdVector3*, unsigned int); // 0x50
         float ambientLight;
-        uint32_t unk;
+        float unk; // 1 / (fov_y/zFar - fov_y/zNear); depth-range normalization scale
```
tim-tim707: rename this field please

louriccia: Done in fc26c39 — named it `depthRangeScale` (the 0x58 float BuildFOV writes as `1/(fov_y/zFar - fov_y/zNear)`). Rebuilt clean.

Side note: the field just above it (`ambientLight` @0x54) also looks mislabeled — BuildFOV stores `nearScale = fov_y/zNear` into it. Left it alone to keep this focused, but happy to rename it too if you want.

PR #249, opened 2026-07-10
Add an opt-in per-profile randomizer (offline)

louriccia:
Adds an offline, per-profile randomizer in the `dinput_hook` delta layer -- the reimplemented `src/` sim stays byte-faithful; this only bridges to the engine at menu/race seams.

**How it works:** opt in when creating a profile. The profile **name is the seed**, and the chosen categories freeze into a sidecar (`data/player/randomizer.ini`), so a run is deterministic, reproducible, and shareable. Off by default = a normal vanilla profile (no sidecar, no effect). Existing profiles are never touched.

**Categories:** track order, pod-handling shuffle, starting truguts, shop prices, race winnings, starting pods (1-23 slider), track pod-rewards, mirror mode, lap count, AI difficulty/spread. Each draws from its own PCG stream so toggles are orthogonal.

Distinct from the Archipelago client -- this is a self-contained single-player toy, no server/multiworld.

Draft for now -- opening it up for a read on whether this belongs upstream and how you'd want it scoped. Small `src/` touches ride along: 3 named profile globals in `data_symbols.syms`, `swrUpgradeInfo.unk4 -> cost`, and `swrUI_VISIBLE` added to the `swrUI_FLAG` enum.


tim-tim707:
needs rebase for me to test this

louriccia:
Rebased onto current master — merged clean, builds and links.

One thing to flag while you're testing: this branch was defining 0x00e35a94 a second time as `swrRace_workingPodUnlockMask`, which master already names `swrRace_favPilotUnlockMask`. I adopted master's name and migrated the single caller, so the symbol checker passes.

That said, the research on this branch reads that address as the general `podracers_unlocked` bitfield at profile +0x34 — which would make `favPilotUnlock` too narrow. I didn't want to rename a canonical symbol inside a rebase, so I left it; happy to do it as its own change if you agree with the wider reading.

PR #253, opened 2026-07-12
feat(renderer): restore N64 reflective texgen (#206)

### LINKED ISSUE #206: "Reflective" Textures
Some materials have a pseudo-reflective effect on N64. This is a somewhat-comprehensive list:

**Pods**
- Anakin's cockpit top and windscreen
- Ark Bumpy Roose's cockpit sphere and cable connectors
- Dud's windscreen?
- Fud Sang's cockpit + engines

**Parts**
- Repulsorgrip green parts
- Control Linkage/Stabilizer/Coupling/Nozzle / Control Shift Plate/Control Vectro-Jet rods
- Dual/Quad32PCX/44 / Mag6 Injector coils
- Plug2/3/5/8 Thrust Coil cylinder ring / Block5/Block6 Thrust Coil nub
- Rod Coolant Pump/Dual Coolant Pump/Turbo Coolant Pump rods
- Power Cell/Plug connectors

**Tracks**
- Vengeance fans

louriccia:
Closes #206.

Restores the N64 pseudo-reflective "chrome" texgen (Ark/Anakin cockpits, Fud Sang engines, upgrade parts, Vengeance fans, ...) that the PC port lost when it baked static UVs and dropped the `G_TEXTURE_GEN` geometry bit.

The GL renderer re-generates sphere-map UVs from the view-space normal (`n64_shader.vert`, gated on a `texgenMode` uniform). The texgen flag itself doesn't survive in the PC data — reflective meshes span several combiner/render-mode configs, some byte-identical to normal lit panels — but they all sample the reflection texture `chrome01` (TEXID 35), which does survive, so that's the gate. It's a heuristic (a chrome01 mesh the N64 didn't texgen would false-positive), but in testing it matched the N64 with no false positives; there's a comment marking where to add a secondary filter if one turns up.

On by default (faithfulness), with a toggle + detail/rotation/offset tuning sliders. Values were eyeballed against N64 reference footage since the real texgen scale isn't recoverable from PC data. Also bundles a small dev-tool upgrade: the texture picker now resolves the exact mesh under the cursor (not just its texture) and shows a hovered-mesh material readout — handy for the sibling N64-combiner issues.


dinput_hook/imgui_utils.cpp
```diff
@@ -1185,6 +1223,24 @@ static void panel_graphics_settings() {
     if (ImGui::Checkbox("Enable fog", &imgui_state.enable_fog)) {
         save_settings_ini();
     }
+    // N64 pseudo-reflection texgen on meshes sampling chrome01 (issue #206).
+    if (ImGui::Checkbox("N64 reflective texgen (#206)", &imgui_state.reflection_texgen)) {
```
tim-tim707: remove the text (#206) here

tim-tim707:
I tested locally but something doesn't seem to work. I don't get reflections on anakin cockpit or ark bumpy cockpit like you showed me last time

louriccia:
Found it — it's the HD renderer, not the texgen.

`swr_draw_mesh` does this before it ever gets to the N64 material code:

```cpp
if (imgui_state.HD_replacement) {
    if (try_replace(model_id, ...) && !imgui_state.show_original_and_replacements) {
        return;
    }
}
```

The texgen uniform is set ~200 lines further down, so a replaced model never reaches it. `HD_replacement` defaults to on whenever `assets/gltf` exists, and both `AnakinSkywalker_pod` and `Ark_Bumpy_Roose_pod` are in `replacement_names.md` — so on your setup the two pods you went looking at are exactly the ones drawn by the glTF path. When I demoed it I must have had HD off. Turning HD model replacement off should show it immediately.

I didn't push texgen into the glTF renderer: those models have their own materials and the env-cubemap reflection path, and this PR is about restoring the *N64* renderer's behaviour. Instead the panel now says so when HD replacement is on, so the toggle doesn't look broken:

> HD model replacement is on: replaced models (pods included) draw through the glTF renderer and never reach this. Turn HD replacement off to see it on pod chrome.

Also dropped the `(#206)` suffixes from the labels as you asked — kept the issue reference in the code comments. Pushed, builds and links clean.

If you'd rather replaced pods got reflections too, that's a real feature but a different one (feeding the env cubemap into the glTF material), and I'd do it as its own PR.

PR #261, opened 2026-07-14
Fix three misleading comments in swrRace physics

louriccia:
Comment-only cleanup in `swrRace.c` after re-verifying the pod flight model against the Ghidra decompile. No behavior change.

- **`swrRace_ApplyGravity`** — negative `pitch` scales the fall step *down* (`* (1 + 0.9*pitch)`), i.e. slower descent / glide. The old comment said "nose-down pitch boost"; it's a reduction, not a boost.
- **`swrRace_UpdateSpeed`** — the `AI && pitch < -0.5` speed bonus (1.3x / 1.9x) isn't a player nose-down input. For AI, `pitch` is driven to `-1.0` by `swrRace_UpdateDeathPitch_Maybe` when the pod glides above its spline target, so this branch only fires in that glide-assist state.
- **`swrRace_MainSpeed`** — `unk1998` is the distance to the nearest local player (set in `swrObjTest_F0`), so `(unk1998 - 400)/600 >= 1` is an LOD cutoff at ~1000 units, not a "race timer past its limit".

Builds/links clean (`dinput.dll`).

### DIFF
--- a/src/Swr/swrRace.c
+++ b/src/Swr/swrRace.c
@@ -1288,7 +1288,8 @@ void swrRace_ApplyGravity(swrRace* player, float* a, float b)
         player->fallValue += swrRace_deltaTimeSecs;
     }
 
-    // fallRate = dt * unk190 * fallValue * 30, with a nose-down pitch boost.
+    // fallRate = dt * unk190 * fallValue * 30. Negative pitch scales the fall step DOWN by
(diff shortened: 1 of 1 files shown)


louriccia:
Folded into #262 (the three comment fixes are now in that PR, translated to the renamed symbols). Closing this in favor of #262.

Issue #30, opened 2025-02-08
Rendering issues on Ando Prime, small rendering issues


tim-tim707:
- [x] Transparent objects must not Z-Write, or funny stuff happen.

- [x] Also, mirror flag for ice is toggled on node 1 0 which should get the proper matrix
- [x] have an imgui flag to show both replacement and original models on top of each others for easier comparison.
- [ ] flash the pod blue when respawning

louriccia:
I am unable to replicate this issue on latest build. Non-replaced pods properly blink blue on respawn on all tracks, including all 4 Ando Prime courses. 

<img width="1391" height="987" alt="Image" src="https://github.com/user-attachments/assets/bcd39591-bb9e-43c2-8254-c389efc03a35" />

tim-tim707:
Sorry for wasting your time, I should have given much more context.

The pod does flash blue even with mods enabled on any map, but it does not with the 3D renderer replacement. This is probably an easy fix but I don't know where to read the blue value / respawn timer value.

Do you have an idea ?

louriccia:
Oh I think I understand now, did you mean 3D model replacement? Like how lego anakin does not flash blue when respawning (and has no cables).

louriccia:
Got it working on the renderer replacement in #127. The flash is a per-pod *light* override (`swrObjTest_F3` sets the pod's light blue while `flags0 & 0x6000`), which the IBL-lit HD pods never read — reproduced it as a tint on the PBR pod gated on the same flag, matching the engine's `frac(timer*4)` blink rate. Player pod for now.

PR #266, opened 2026-07-15
Render second bank light (wall-scrape spark illumination)

louriccia:
On N64, when a pod scrapes a wall the sparks light up that side of the pod. That effect is actually still driven in our code and just wasn't being rendered.

The game gives each pod its own light bank and, while scraping (`flags0 & 0x30000000`, the scrape-spray flags), `swrObjcMan_UpdateLighting` fills that bank's **second** light (`lightColor2`/`lightDirection2`) aimed from the spark contact point toward the pod and flickers it. The vanilla D3D path renders it via `rdCamera_SetLights` (activates `lights[1]` when `numEnabledLights > 1`), but our GL mesh path only ever uploaded bank light 1 -- there was a literal `// TODO light 2` where the second light should go.

This uploads the second bank light when `num_enabled_lights > 1`, matching `rdCamera_SetLights`. The secondary bank has no ambient term, so it's diffuse-only. Same machinery also backs the respawn/death flash's light path.

Scope: native GL mesh lighting only. **HD-replacement pods won't show it** yet -- they're lit by IBL and ignore the banks (same limitation as the #30 respawn flash that needed `podTintColor`); that's a separate follow-up.

Tested: wall-scrape on both engine sides lights the correct side; ordinary scenes unchanged (banks stay at 1 light everywhere else).


tim-tim707:
Do you have a screenshot for comparison ? I struggle to see a difference

louriccia:
there is a subtle strobe light effect when scraping walls

with second bank light:
image

without:
image

PR #271, opened 2026-07-16
perf(renderer): frustum culling, GL state dedup, streamed uploads (16 racers: 50-60 -> ~110 fps)

louriccia:
Follow-up to #166. The GL renderer drew the whole scene graph with no visibility testing, so framerate scaled badly with racer count: 15 off-screen pods cost ~1400 full-price meshes per frame, ~94% re-parsed + re-uploaded every frame (pod part matrices animate even at idle). Profiled with Tracy on a Release build (16-racer grid, vanilla N64 pods); the Tracy scaffold isn't included here to keep the diff focused.

| full-field-on-screen, per frame | before | after |
|---|---|---|
| render_mesh CPU | 8.2 ms | 2.7 ms |
| vertex upload | 2.7 ms | 0.23 ms |
| draw submission | 2.6 ms | 0.76 ms |
| GPU scene | 6.3 ms | 0.39 ms |
| **fps** | **50-60** | **~110** (field off-screen ~120+, menus 200+) |

- **Frustum culling**: local-space mesh AABB vs clip volume; culled meshes skip state/upload/draw. Skinned meshes (`vertex_base_offset != 0`) are never culled (their vertices come from other meshes' parses), and a culled mesh still parses whenever the drawn path would have, so the N64 shared-vertex staging stays byte-identical (soft skinning intact). The HD pod path culls the whole pod up front from its engine/cockpit origins.
- **GL state dedup**: shadow what the mesh path last set (render mode / program / texture + params / cull face / every uniform per shader) and skip unchanged calls. Also removes the per-mesh `glUseProgram(0)`+VAO unbind that forced a full rebind on every draw.
- **Streamed uploads**: animated meshes memcpy into a persistent-mapped ring buffer (4 fenced regions, one per presented frame) and draw from a vertex offset instead of a per-mesh `glBufferData`. Falls back to the old path on region overflow or missing GL 4.4 buffer storage.
- **HD env-cubemap fix**: the flag ending per-mesh env-to-cubemap stamping was nested inside a debug gate, so a normal HD race stamped every mesh into the 2048² reflection cubemap every frame (~11 ms). Now unconditional -- race pods reflect the skybox, with a "HD pod scene reflections (slow)" toggle to opt back in until captures can be precalculated per track.
- **Free debug groups in Release**: `PushDebugGroup(std::format(...))` was building thousands of throwaway strings per frame for a compiled-out GL call; now a no-op macro.

Each optimization has a Graphics Settings toggle (default on). Also adds a runtime **VSync toggle** (was hardcoded on at GL open) -- on a 60 Hz display none of these gains are visible without it.

### DIFF
--- a/dinput_hook/gltf_utils.cpp
+++ b/dinput_hook/gltf_utils.cpp
@@ -75,17 +75,16 @@ void loadGltfModelsForTestScene() {
     }
 }
 
-void PushDebugGroup(std::string message) {
+// In Release these are macros (see gltf_utils.h) so call sites don't even build their argument.
 #if !defined(NDEBUG)
+void PushDebugGroup(std::string message) {
     glPushDebugGroup(GL_DEBUG_SOURCE_APPLICATION, 0, message.length(), message.c_str());
-#endif
 }
 
 void PopDebugGroup(void) {
-#if !defined(NDEBUG)
     glPopDebugGroup();
-#endif
 }
+#endif
 
 unsigned int getBufferByteSize2(const fastgltf::Accessor &accessor) {
     return accessor.count * fastgltf::getElementByteSize(accessor.type, accessor.componentType);
--- a/dinput_hook/gltf_utils.h
+++ b/dinput_hook/gltf_utils.h
@@ -160,8 +160,16 @@ void setTextureParameters(GLint wrapS, GLint wrapT, GLint minFilter, GLint magFi
 const std::byte *getBufferPointer(const fastgltf::Asset &asset, const fastgltf::Accessor &accessor);
 
 void loadGltfModelsForTestScene();
+#if defined(NDEBUG)
+// True no-ops in Release: the macro discards its argument unevaluated, so hot call sites like
+// PushDebugGroup(std::format(...)) -- one per mesh/group per frame -- stop paying the string
+// construction that the compiled-out glPushDebugGroup would have consumed.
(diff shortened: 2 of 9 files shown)


louriccia:
Pushed one more commit for the HD-replacement path (same 16-racer test, HD pods on, was ~30 fps):

- `renderer_drawGLTFPod` now frustum-culls the whole pod before doing any work, so off-screen HD pods are free like the N64 ones.
- Found the bigger one with Tracy: the `environment_models_drawn` flip was accidentally nested inside the `show_replacementTries` debug gate, so a normal HD race stamped every mesh into the reflection cubemap every frame (~11 ms/frame of per-mesh FBO churn). Now it flips at the first non-env model as intended -- hangar reflections keep working, races skip the stamping.

Tradeoff: race pods now reflect the skybox instead of live scene captures. There's a "HD pod scene reflections (slow)" toggle (default off) to opt back into the old behavior. Longer term your idea of precalculating the captures per track (e.g. from spline positions) seems like the right way to get the look back without the cost -- deferred for now.

louriccia:
Rebased to drop the vendored Tracy profiler and its instrumentation -- kept the diff to just the perf work + the runtime VSync toggle (9 files, 2 commits). All the numbers above came from Tracy on a local Release build; happy to open the profiler scaffold as its own separate PR if you'd find it useful in-tree.

PR #264, opened 2026-07-14
Name verified swrObjTest fields, flag bits, and functions (follow-up to #262)

louriccia:
Follow-up to #262, same theme: names verified by a full decompile sweep of the swrObjTest orbit (every constant read from the binary, evidence in the commit messages).

Highlights:
- swrRace +0xac is a `swrSplineCursor`, not a matrix - it's the pod's track-spline cursor (typedef moved above swrRace).
- Fixes the long-standing swrRace sizeof bug (0x1f24 vs 0x1f28): every label after 0x1eb8 was 4 bytes off and one trailing dword was missing. Offsets are now compiler-verified (`_Static_assert` checked locally).
- The "spline progress values" scratch table actually stores baked track-mesh pointers: +0xec/+0xf0 are the current/previous track mesh under the cursor, `swrModel_Mesh` +0x3c is that mesh's spline node index, and `swrRace_UpdateSplineBinding` uses it to re-seek the cursor to the mesh the pod is standing on.
- swrScore +0x18 points at the racer's pilot id (0..22). That unlocked the pilot specials: pilot 2 = Sebulba's flame attack (the `SpawnExplosionEffect` / `SendFlameEvent` / `SpawnExplosionByIndex` trio is that whole networked mechanism, renamed to `*FlameAttack*` - it's not the death explosion), pilot 14 = Neva Kee's fused engines (cockpitXf copied onto both engineXf in F3).
- flags0's low nibble is a state field (0 pre-race / 1 post-race / 2 racing); new named bits: REPAIRING, WALL_IMPACT_CLAMP, LOOK_BACK, GUIDE_ARROW (kills the raw 0x8000000 in swrObjJdge_UpdatePlayerHUD), SCRAPE_SPARK_R/L, THROTTLE_SETTLED, plus flags1 RESPAWN_RELIGHT. The two _Maybe bits from #262 are superseded - 0x100000's missing consumer is the camera manager's look-back flip.
- ~20 functions drop `_Maybe` or get corrected names; the notable misnames: `UpdateScrapeContact` is really a Smok-geyser proximity push, `setInRaceInputState` is a raycast ignore-facing toggle, `FxAdvanceAnimTimers` applies the pod's vertical part sink offset.

Full dinput_hook build + link green; dup/data-symbol/hook-annotation scripts clean; my Ghidra DB renamed to match.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #273, opened 2026-07-19
reimpl(times): lap/total timing, HUD/results display, save-game persistence

louriccia:
Reimplements the whole times pipeline: lap/total time accumulation (`swrObjJdge_F2` + the lap-progress chain), the in-race HUD timer and post-race results display, the full elfSaveLoad save/profile block (tgfd.dat + .sav exports, CRC32, defaults) - and, in the second commit, the two menus that close the loop: `swrRace_ResultsMenu` (where new track records actually get committed: holder name + pilot into the save image, plus the whole tournament progression step - circuit-scaled truguts, track/circuit/invitational unlocks, beat-place bits, favorite-pilot unlock, podium hand-off) and `swrRace_CourseInfoMenu` (option rows + records display on course select).

**Stacks on #264** - merge that one first, bottom-up; the first seven commits here are its. This diff narrows to the times work once it lands.

Restacked onto it because the two PRs were duplicating groundwork: both had independently added the *identical* 12-line `swrSplineCursor` typedef and the same `swrRace_GetTrackMeshAtCursor_ADDR` define, and this branch's reimpl bodies were written against the pre-#264 field names. Merged independently that meant a types.h conflict plus a duplicate typedef definition; stacked, #264 owns the naming and the `chore(times): adopt #264's canonical names` commit here shows exactly what moved (`unkec_node` -> `splineTrackMesh`, `unkf4/f8/fc/100` -> `splineProjOut1/Result1/Out2/Result2`, `unk10c`/`unk10e` -> `checkpointCount`/`splineRebindResult`, `swrScore.unk18` -> `int* pilotId`, plus the de-`_Maybe`'d swrSpline stub names). #264 picked up one commit from this pass too: `swrRace+0x128` is now `gapToLeader`, matching the sibling gap fields #282 names at 0x12c..0x134.

The fun find: lap timing is sub-frame precise. `swrRace_UpdateRaceProgress` reports the slice of the frame delta spent before the line crossing; F2 closes the finished lap with that and opens the new lap with the remainder (using the raw, unclamped delta, capped at 3000s).

Struct corrections that fell out of the disassembly (all static-assert verified):
- `swrScore.results_P1_Lap` (0x78) is an **int** - F2 does an integer `inc` and GetRacerProgress `fild`s it. It was declared float, so the existing `(float)(int)` casts were latently wrong; fixed all users incl. the 100-lap delta.
- `swrRace` had a 4-byte hole at 0x1ef8, so `unk1f00..unk1f24` sat 4 bytes low and sizeof was 0x1f24. Restored to 0x1f28; the existing unk1f00/1f18/1f1c refs now hit their true offsets.
- `swrObjJdge` 0x134 isn't a matrix: it's a `swrSplineCursor postRaceCursor` + `rdMatrix44 postRaceMat` pair (the old "unk1a0 = 1.0" is just the matrix's vD.w, and F0's finish->results gate turns out to be `postRaceCursor.endFlag`).
- tgfd.dat now has a typed layout (`swrSaveData`/`swrSaveProfile` in types.h): 4 profile slots + per-track record tables indexed `[track*2 + mirror]`, defaults 3599.99s / 32x'A' / per-track favorite pilot. `swrSaveProfile` +0x24 decoded: [0] is the profile's selected **pilotId** (what gets stamped as record holder), [1..3] per-circuit tracksUnlocked bitmasks, [4] circuitsCompleted.
- `swrObjHang_STATE` 16/17 identified (PODIUM, PILOT_UNLOCK), and `jdge->hud_mode` got a proper `swrObjJdge_HUDMODE` enum (gap arrows / progress ring / minimap far+near / off / splitscreen column variants).
- `swrObjHang_StartRace` documented: the 'Begn' payload carries the selected track's saved records, which `swrObjJdge_F4` latches into `jdge->best_lap_time_ms` / `recordLap3_ms` - closing the loop on how records flow save -> race -> HUD -> results -> save.

Two `_Maybe`s resolved along the way: `swrRace_LoadProfile` (pairs with SaveProfile) and `swrRace_DrawRecordHolderName`.

Also adds dormant HANG stubs for the not-yet-reimplemented callees so the DLL links, and a new `swrWeather.c` for the weather setter stubs. Full build green after each commit, re-verified with a full dinput.dll link on the stacked branch.


tim-tim707:
this also need a rebase

louriccia:
Rebased onto current master (merged it in, per this repo's idiom) — builds and links clean.

Three conflicts, all resolved toward the names that landed on master:
- `swrObjJdge` 0x134/0x164 → `camSweepCursor` / `camSweepTransform` from #301, over this branch's `postRaceCursor` / `postRaceMat`.
- `swrRace` 0x2b8 → `unk2b8`, the force-feedback vibration magnitude from #303. This branch had separately found that `swrRace_InRaceTimer` decays it, so I folded that into the comment — the two readings agree.
- `data_symbols.syms` 0x4c4000 → kept `swrObjHang_menuJustEntered` over master's `DAT_` placeholder, alongside master's two new swrObjHang camera symbols.

While in there I also fixed something the symbol checker was failing on before the rebase: this branch defined two addresses twice.
- 0x004c022c — `swrRace_localPlayerPlace` → `swrRace_resultsPlaceP1`
- 0x0050c540 — `swrRace_resultsMilestones` → `swrRace_resultsStateFlags`

In both cases the survivor is the better-researched name and carries ~10 references against 1 for the one I dropped; the two delta-layer callers moved over, and the raw `|= 8` in `swrRace_ResultsMenu_delta` is now `swrRace_RESULTSFLAG_PILOT_UNLOCK_SHOWN`.

PR #274, opened 2026-07-19
reimpl(times): results + course-info menus (track-record commit and display)

louriccia:
Stacked on #273 (its commit is included here; merging that first shrinks this diff to the second commit).

This finishes the times pipeline: `swrRace_ResultsMenu` is where track records actually get committed — it sanitizes and sorts the race scores, sends record-setting guest profiles through name entry, writes new 3-lap / best-lap records into the save image (holder name + pilot from the live profile), applies the full tournament progression step (circuit-scaled prize truguts, next-track/circuit/invitational unlocks, per-track beat-place bits, the favorite-pilot unlock cutaway, podium hand-off), and autosaves. `swrRace_CourseInfoMenu` is the other end: option rows, track preview, and both saved records with holder name + pod sprite.

Decodes that fell out:
- `swrSaveProfile` +0x24 isn't generic unlock data: [0] is the profile's selected **pilotId** (that's what gets stamped as record holder), [1..3] are per-circuit tracksUnlocked bitmasks, [4] is circuitsCompleted.
- `swrObjHang_STATE` 16/17 identified: PODIUM (top-3 pilots in `podiumCharacters`) and PILOT_UNLOCK.
- `swrObjHang_StartRace` documented: the 'Begn' payload carries the selected track's saved records, which `swrObjJdge_F4` latches into `jdge->best_lap_time_ms` / `recordLap3_ms` for the session — closing the loop on how records flow save -> race -> HUD -> results -> save.
- Named the results/course-info globals and swapped tracks_delta's raw `DAT_` uses over to them.

Full build green.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -358,6 +358,7 @@ swrRace_DeathSpeedDrop 0x004c7bbc float = 140.0
 swrRace_FireTimer 0x004c7bc0 float = -1.0
 
 swrSpline_trackLength 0x004c7be0 float
+swrSpline_finishNodeIdx 0x004c7be4 int # -1 closed-loop track, -2 open spline (CP0 has no prev; set by swrSpline_Build), >= 0 open-track finish: lap progress completes at 1.0 (swrRace_LapProgress / LapCompletion, GetCustomStartTransform)
 
 sithControl_secFPS 0x004c8174 float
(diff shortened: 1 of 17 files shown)


louriccia:
Folding this into #273 as a second commit instead of a stacked PR.

PR #281, opened 2026-07-22
reimpl(swrRace): swrRace_UpdatePlayerControl

louriccia:
Reimplements `swrRace_UpdatePlayerControl` (0x46bec0) — the per-frame human-input -> pod-intent routine.

**Stacks on #264** (#280 has merged) — the first seven commits here are #264's; please review/merge that first, bottom-up. Review only the last three: `f23cb33` (the reimpl), `cfb4fde` (the 0x220/0x224 docs answering the review question below), and `f4249df` (the name retarget). This diff narrows to those once #264 lands.

Restacked onto #264 because it owns the swrObjTest/swrRace naming and this branch was written against the older names: it renames `unk1eb8`/`unk12_1`/`unk4_mat`/`unk2f4`/`unk2f8` and `score->unk18`, and both PRs had independently named the 0x220/0x224 pitch pair. Kept independently and merged in either order, that meant a types.h conflict plus a `swrRace.c` that compiles against neither side alone. Stacked, #264 owns the names and `f4249df` shows exactly what moved. Two collisions went #264's way because it identified a consumer this branch could not — `0x100000` is `FLAG0_LOOK_BACK`, and `0xac+0x28` is `splineCursor.branchSelector` (so the script 5/6 block is a spline-fork override, not a camera flag); `0x400000` keeps this branch's `BOOST_OVERDRIVE`, which #264 left unnamed.

Control-type dispatch (analog / per-player bitset / debug demo-replay), engine-damage steering pull + force feedback, boost-start detection, flame-attack/taunt, then the turnRateTarget/throttle/tilt/flags commit. Reverse-hooked (dormant).

- Names the swrControl per-action input globals (button-state `@0x00ec8840[]` / hold-time `@0x00ec88a0[]` elements) + the debug/FF latches; fixes `DAT_00e98e80/ea0` `float[1]`->`float[4]`; retypes `swrRace_DebugFlag` char->int (retail reads it as a 32-bit dword — bits 0x100/0x2000000 were dead as a char, which also silently broke an existing `& 0x2000` test).
- `swrObjTest_FLAG0_BOOST_OVERDRIVE` (0x400000): overdrive throttle floor, read here + cleared by `swrRace_BoostCharge`; never set in retail (cut boost companion — read+cleared only).
- Tuning constants -> `engine_config.h` (`SWR_CTL_*`).
- Preserves the retail bug where the analog pitch low-clamp writes +0.8 (not -0.8).
- HANG("TODO") reverse-hook stubs for the bodyless callees so the dll links (swrControl_PlayForceEffect/StopForceEffect, swrMultiplayer_SendEvent, swrRace_CheckResetInput/ApplyEngineDamage/SpawnExplosionEffect/SendFlameEvent_Maybe).

Builds + links clean. `DAT_00e98e80/ea0/ec8854` are left as-is (shared with the delta-layer `tracks_delta.c`; renaming would be a separate cross-layer change).

### DIFF
--- a/src/Swr/swrRace.c
+++ b/src/Swr/swrRace.c
@@ -826,7 +826,7 @@ void swrRace_SetAngleFromTurnRate(float* out_tilt, float cur_turnrate, void* unu
 }
 
 // 0x0044afb0
-void swrRace_GetEngineNodeOffsetPos_Maybe(void** nodePair, rdVector3* outPos)
+void swrRace_GetEngineNodeOffsetPos(void** nodePair, rdVector3* outPos)
 {
     if (nodePair == NULL) {
         rdVector_Set3(outPos, 0.0f, 0.0f, 0.0f);
@@ -852,7 +852,7 @@ void swrRace_GetEngineNodeOffsetPos_Maybe(void** nodePair, rdVector3* outPos)
 }
 
 // 0x0044b270
-void swrRace_SetEngineNodeTranslation_Maybe(void** nodePair, rdVector3* pos)
+void swrRace_SetEngineNodeTranslation(void** nodePair, rdVector3* pos)
 {
     if (nodePair == NULL)
         return;
@@ -872,7 +872,7 @@ void swrRace_SetEngineNodeTranslation_Maybe(void** nodePair, rdVector3* pos)
     }
     rdMatrix44 offsetMat;
     swrModel_NodeGetTransform(offsetNode, &offsetMat);
-    // remove the engine-height offset applied by swrRace_GetEngineNodeOffsetPos_Maybe
+    // remove the engine-height offset applied by swrRace_GetEngineNodeOffsetPos
     rdVector_Scale3Add3((rdVector3*) &out.vD, (rdVector3*) &out.vD, -offsetMat.vD.y, (rdVector3*) &out.vB);
     swrModel_NodeSetTransform(node, &out);
 }
--- a/src/Swr/swrRace.c
+++ b/src/Swr/swrRace.c
@@ -1585,20 +1585,20 @@ float swrRace_RaycastGround(swrRace* player, rdVector3* pos, int* outSurfaceNorm
 
     swrRace_ResetCollisionHit();
     if ((player->flags1 & swrObjTest_FLAG1_FULL_RAYCAST) == 0)
-        hitDist = swrModel_CollideRayWithMesh((swrModel_Mesh*) player->unkec_node, ray,
+        hitDist = swrModel_CollideRayWithMesh((swrModel_Mesh*) player->splineTrackMesh, ray,
                                               (float*) &outPoint, (float*) &outNormal);
     else
         hitDist = -1.0f;
 
     if (hitDist < 0.0)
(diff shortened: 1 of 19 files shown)


src/Swr/swrRace.c
```diff
@@ -1127,6 +1132,524 @@ void swrRace_UpdateCatchup(swrRace* player)
...
+    if ((player->flags0 & swrObjTest_FLAG0_ZON) != 0) {
+        if (pitchForward >= SWR_CTL_STEER_DEADZONE || -pitchForward >= SWR_CTL_STEER_DEADZONE) {
+            float squared = pitchForward * SWR_CTL_STEER_SQUARE_GAIN * (pitchForward * SWR_CTL_STEER_SQUARE_GAIN);
+            if (pitchForward < 0.0f) {
+                squared = -squared;
+            }
+            player->unk10_2 = -(player->podStats.maxTurnRate * squared * SWR_CTL_STEER_SCALE) * SWR_CTL_HALF;
+        } else {
+            player->unk10_2 = 0.0f;
+        }
```
tim-tim707: if we know what unk10_2 does can you rename it accordingly ?

tim-tim707 (changes requested):
there seem to be some fields and flags that could be renamed better

src/Swr/swrRace.c
louriccia: Traced it -- `unk10_1`/`unk10_2` (0x220/0x224) are a current/target pair for the pod's zero-g **pitch rate**, the pitch-axis twin of `turnRate`/`turnRateTarget`. Renamed to `pitchRate` / `pitchRateTarget` in 84782cf.

The loop, all three stages gated on `FLAG0_ZON`:
1. here in `UpdatePlayerControl`, the pitch stick writes the target scaled by `podStats.maxTurnRate`, then zeroes the normal pitch path
2. `swrObjTest_UpdateControlAndMove` slews 0x220 toward 0x224 at `podStats.turnResponse * multiplayerStats * dt` -- the same rate expression `swrRace_UpdateTurn` uses for `turnRate -> turnRateTarget`
3. `swrRace_UpdateSplineOrientation` applies 0x220 (x1.5) as a rotation about the pod's local X axis

Stage 3 is what nails the axis, since it applies all three side by side:

| axis | field | |
|---|---|---|
| `(0,0,1)` -> `vC` (up) | `turnRate` @0x1ec | yaw |
| `(1,0,0)` -> `vA` (lateral) | **0x220** | **pitch** |
| `(0,1,0)` -> `vB` (forward) | `gravityTubeAngle` @0x21c | roll |

`vB` is forward (the speed-sign dot product uses it) and `vC` is up (gravity is `-vC`), so `vA` is lateral -> pitch. The neighbours were already named for the other two axes, which corroborates it. Makes sense in-game too: in zero-g the stick pitches the pod instead of trimming it.

Left `unk10_3` @0x228 alone -- `UpdateSurfaceTag` writes 3.0f on the ZON->ZOFF transition and `Init`/`ResetToSpline` clear it, but nothing in the binary reads it, so I'd only be guessing. Happy to comment it as dead if you'd prefer.

Also rebased onto current master (#280 merged, so its commit dropped out -- this PR is down to the 2 real commits). Full build links green.

PR #284, opened 2026-07-31
fix(math): take float constants from the stored bits, not the decompiler's text

louriccia:
Stacked on #283 — that one first.

Ghidra prints a float32 constant as a shortened decimal that doesn't always round-trip, and the reimplementations copied that text without an `f` suffix. You get three different numbers from one constant. At `0x004adf9c` the stored value is `0x42055556` = 33.333335876464844; Ghidra prints `33.33334`; but `float("33.33334")` is `0x42055557`, one ULP off — and as a bare double literal it's a third value again. The correct literal is `33.333336f`.

Two of these weren't roundings but **branch decisions**:

- `stdMath_ArcTan2`'s thresholds are float32 `9.999999747e-05`, strictly *less* than the double `0.0001`. At `x2 == 0.0001f` retail compares equal and evaluates the polynomial, while the double compares greater and takes the 90-degree early-out — a 751 ULP divergence.
- `rdMatrix_BuildFromVectorAngle44` compares against float32 `0.999f`; the bare double `0.999` is a slightly smaller number.

Also fixed coefficients that were printed lossily (`-0.16666667f` not `-0.1666667`, `0.70710677f` not `0.7071068`, `57.295784f` not `57.29578`, and friends). Constants genuinely stored as doubles — 1/pi at `0x004ac6a0`, pi and 1/180 in `SinCos` — are left as doubles.

With this applied, all of stdMath, rdVector, rdMath and rdMatrix reproduce retail **bit-for-bit** over several hundred thousand sampled inputs each. `Decelerator` alone went from differing on 19835 of 20000 inputs to exact. No ULP tolerance needed anywhere.

Worth knowing for future reimpl work: the decompiler's float text is not safe to copy. Reading the constant's bytes and emitting a literal that round-trips is the reliable move.

### DIFF
--- a/src/General/stdMath.c
+++ b/src/General/stdMath.c
@@ -24,8 +24,19 @@ void stdMath_MultiplyAddClamped(float* res_inout, float value, float multiplier,
 }
 
 // 0x00429d90
-void stdMath_AddScaledValueAndClamp_i32(int* res_inout, int value, float multiplier, int min, int max)
+void stdMath_AddScaledValueAndClamp_i32(int* res_inout, float value, float multiplier, int min, int max)
 {
+    // Retail is FLD multiplier / FMUL value / FILD *res_inout / FADDP / __ftol, so the
+    // whole accumulate happens on the x87 stack: the product is formed at register
+    // precision, the running total arrives exactly via FILD, and only the final result
+    // is truncated toward zero. Double intermediates reproduce that -- float ones would
+    // round the product down to 24 bits of mantissa before the add.
+    //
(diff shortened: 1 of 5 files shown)


louriccia:
Folding this into #283 instead. These constant fixes land in the same functions as the behaviour fixes there — `QuaternionToAxisAngle` had its typo in one PR and its thresholds in the other, so you'd have been reading half the story in each. They're also textually adjacent in `rdMath.c`, so splitting them made the diffs overlap rather than narrow.

#283 now carries both and is the only prerequisite for #285. Nothing lost — same content, one coherent read.

Issue #277, opened 2026-07-20
Triage


louriccia:
_Last swept 2026-09-06 (second sweep of the day)._ **Five more PRs merged since the earlier 09-06 sweep: #291 (crash-report build identity + build stamp in the debug menu header), #273 (times pipeline), #304 (F12 screenshot), #253 (N64 reflective texgen) and #215 (collision viewer).** Burndown: **945** fully-reimplemented bodies (up from 917), 2131 of 2137 named (99.7%); only **6** `FUN_` functions remain unnamed in the whole DB -- the naming campaign is effectively done.

**#299 is CLOSED as superseded.** #273 reimplemented `swrRace_CopyProfileFromSave` (0x0044e500) and `swrRace_CopyProfileToSave` (0x0044e530) at the same addresses, typing `swrRace_aProfiles` as `swrSaveProfile[20]` and doing a plain struct assignment where #299 still used `memcpy` with hand-rolled offset constants. Its stack parent #298 is unaffected.

**SCOPE NOTE -- the #219 split is now half done.** tim's ask on #291 was: _"I don't want to merge #219. We should make an internet data registry that could be toggled on and off with a single button, with cache, for speedrun records, version check and more. Assume we never get any data from the registry and have default values for every key. So we should split 219 into the internet registry PR and the version display."_ **Half (b), the version/build display, shipped with #291** -- the build stamp is in the debug menu header and is click-to-copy. Half (a), the offline-safe internet data registry (links overlay, update check, speedrun records; one toggle, cached, default value for every key) is still to be built, and is what remains of #219.

**Rebase state: clean.** Every open PR now merges cleanly against master except **#134** (draft prototype, 464 behind, 7 conflicting files across `swrControl_delta` / `swrObjJdge_delta` / `swrRace_delta` / `imgui_utils` / `renderer_hook` -- a rewrite, not a merge). The whole open queue was resynced and rebuilt against the post-merge master this sweep: #302, #298, #276, #265, #260, #249, #219, #217, #182. CI is green on all of them.

This issue organizes all remaining open items from https://github.com/tim-tim707/SW_RACER_RE/issues/143, https://github.com/tim-tim707/SW_RACER_RE/issues/247, and feedback from the latest multiplayer sessions (Jul 19, Jul 27) and points them to existing PRs.

## General
- [x] **Display commit hash/version** on-screen/top of debug for version check + patch check
[shipped] https://github.com/tim-tim707/SW_RACER_RE/pull/291
[...]

PR #291, opened 2026-08-29
Crash logger: stamp the build a report came from

louriccia:
Crash and hang reports carry no build identity, so a report shared by a player can only be tied to a binary by resolving its `DINPUT.dll` offsets against candidate DLLs one at a time. That is slow and genuinely ambiguous — the same offset can land inside the same function in two different builds, and telling them apart needs instruction-level disassembly at the reported offset.

Hit this triaging a 2026-08-29 multiplayer session: 20+ reports across three different builds, and working out which report came from which binary took two symbol tables and a disassembly pass before any actual debugging could start.

Adds a `build:` line next to the existing environment line in both report headers:

```
*** unhandled exception: code=0xc0000005 addr=... ***
  environment: native Windows
  build: fix/mp-leave-crashes @ c84840e (pe 68b1c2a3)
  last stage: running (first frame)
```

- The git revision names the source.
- The PE timestamp, written by the linker, additionally separates two builds of the same commit and still identifies a binary when git was unavailable at build time (revision `unknown`).
- A build with uncommitted changes to tracked files is marked `-dirty`, since the revision alone would not describe it.

The stamp is regenerated on every build rather than at configure time so it cannot go stale after a commit; the generator writes through `configure_file`, which leaves the header byte-identical when nothing changed, so this does not force a rebuild. Missing git degrades to `unknown` instead of failing the build.

Split out of #276, where it was written — it is unrelated to the multiplayer fixes and useful on its own.

### DIFF
--- a/cmake/GenerateBuildId.cmake
+++ b/cmake/GenerateBuildId.cmake
@@ -1,7 +1,6 @@
-# Regenerates build_id.h with the current git revision. Run at BUILD time (not configure time)
-# so the stamp cannot go stale after a commit. Writes through configure_file, which leaves the
-# file untouched when the contents are unchanged, so this does not force a rebuild every time.
-#
+# Regenerates build_id.h with the current git revision. Runs at BUILD time, not configure time,
+# so the stamp cannot go stale after a commit; configure_file keeps the header byte-identical when
+# nothing changed, so this does not force a rebuild.
 # Expects: SRC_DIR (repo root), IN_FILE (template), OUT_FILE (generated header).
 
 set(SWR_BUILD_COMMIT "unknown")
@@ -32,7 +31,7 @@ if(GIT_FOUND)
         set(SWR_BUILD_BRANCH "${_branch}")
     endif()
 
-    # Uncommitted changes to tracked files: the stamp alone would not describe the binary.
+    # Uncommitted tracked changes: the revision alone no longer describes the binary.
     execute_process(
         COMMAND "${GIT_EXECUTABLE}" diff --quiet HEAD
         WORKING_DIRECTORY "${SRC_DIR}"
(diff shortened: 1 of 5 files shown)


tim-tim707:
Can you rebase this please ?

louriccia:
Rebased onto current master (merged it in, per this repo's usual idiom) — builds and links clean, and CI is running now.

tim-tim707:
Now that I think of it, and since we embed some git informations, it would be good to also display them in the imgui menu

louriccia:
> it would be good to also display them in the imgui menu

I'm pretty sure that's what https://github.com/tim-tim707/SW_RACER_RE/pull/219 does

tim-tim707:
yes but I don't want to merge #219 .

We should make an internet data registry that could be toggled on and off with a single button, with cache, for speedrun records, version check and more. Assume we never get any data from the registry and have default values for every key.

So we should split 219 into the internet registry PR and the version display

louriccia:
Done — the build stamp now shows in the imgui menu, in the `SWE1R Debug` window header under the FPS line:

```
build: feat/crash-logger-build-id @ 9011aaac-dirty
```

Click it to copy. It's the same string the crash reports carry, built from the `build_id.h` this branch already generates, so there's one mechanism for the embedded git info rather than two — #219 had grown a second one (`generate_git_version.cmake` → `git_version.h`), which is part of why splitting it is the right call.

On the split: agreed, and I've recorded it on #277 so it doesn't get re-litigated. The way I read your shape —

- **this PR** takes the version-display half (done above), and
- **a new PR** does the internet data registry: one toggle, cached, a default value for every key, and it behaves correctly assuming the registry never answers. Version check, speedrun records and the mode-select Discord/SRC links all become keys in it rather than three separate bits of code.

#219 then closes rather than merges, since neither half survives in its current shape.

Before I build the registry — two things I'd rather get right up front than guess at:
1. **Where does it fetch from?** A static JSON in the repo (say `gh-pages` or raw on master) is the least infrastructure and easy for you to edit; anything live means hosting. I'd default to the static file unless you have somewhere in mind.
2. **When does it fetch?** Once on launch in the background, cache to disk with a TTL, never block the game thread — and on a cold cache with no network, every key just reads its compiled-in default. Sound right?

PR #293, opened 2026-09-02
reimpl(swrObjHang): front-end holo-camera family + name its global cluster

louriccia:
Reimplements the six functions that drive the hangar / galaxy-map camera, and names the globals they share.

| fn | addr |
|---|---|
| `swrObjHang_EaseSine_Maybe` | `0x0045a420` |
| `swrObjHang_IsCameraMoving` | `0x00440b50` |
| `swrObjHang_UpdateHoloBillboardMatrix` | `0x0043e210` |
| `swrObjHang_SetHoloCameraTarget` | `0x0045c010` |
| `swrObjHang_BeginCameraMove` | `0x0045c9d0` |
| `swrObjHang_ComputeCameraEye` | `0x0045cb80` |

### The globals are two parallel matrix triples

Not a loose cluster -- one triple for the eye, one for the look-at, each with the current value, the snapshot taken when a move begins, and the value the move ends at. `swrObjHang_LerpHoloCamera` interpolates current from start toward target on both rows, which is what pins the roles:

| role | eye | look-at |
|---|---|---|
| current | `cameraMatrix` `0xe298c0` | `holoCameraMatrix` `0xe2af60` |
| move start | `cameraMatrixStart` `0xe2ae80` | `holoCameraMatrixStart` `0xe2b3e0` |
| move target | `cameraMatrixTarget` `0xe29b60` | `holoCameraMatrixTarget` `0xe2b440` |

### Three symbols were mis-symbolized views into those triples

- **`PodHandlingData3` `0xe2af40` is spurious.** `PodHandlingData` is `0x3c` bytes, and its only reader -- `&PodHandlingData3.heatRate`, at `+0x20` -- lands on `0xe2af60`, which is the `rdMatrix44_unk6` that was sitting commented out in `data_symbols.syms` with a `// conflict with above` note. So the conflict is resolved rather than skirted, and a `rdMatrix_Copy44` that was reading 0x40 bytes out of a 0x3c-byte struct now has a real symbol.
- `rdVector3_unk1` `0xe29b90` is `cameraMatrixTarget.vD` (`0xe29b60 + 0x30`).
- `DAT_00e2af90` is `holoCameraMatrix.vD` (`0xe2af60 + 0x30`).

Also named `swrObjHang_cameraMoveMode` / `Parity` / `Active`, `swrObjHang_sceneCameraEye`, `swrObjHang_wattoCounterEyePos`.

### Two things worth a look

**`cameraMoveMode` `0x0050c930` is typed `short`, not `int`.** Every access in the binary is 16-bit -- reads are cast to short, writes store only the low half -- so `int` would make a faithful write impossible and would falsely claim `0x0050c932`.

**`swrUICameraPlacement` had no typedef**, so the generated `globals.h` accessor for `maybeUICameraPlacements` would not compile from C. Typedef'd it and documented `position1`/`position2` as eye/look-at.

`tracks_delta.c` referenced four of the renamed globals and is updated to match.

### Notes

The family works in degrees (`stdMath_SinCos` takes `angle_degrees`), which is what the +-90 / +-180 / +-360 wrap constants in `UpdateHoloBillboardMatrix` are. The `1.0f / 2147483648.0f` literals are the exact bit pattern of the `0x30000000` the binary uses to map `swrUtils_Rand` onto `[0, 1)`.

The camera transition-mode ids (`mode == 3`, `cameraMoveMode != 5`) are left as literals -- I only see 0/1/3/5 used and did not want to invent enum names for values I have not actually pinned down.

Builds and links clean; hook-annotation, header-dup and data-symbol-dup checks all pass. Burndown goes 889 -> 895 fully reimplemented.

Heads-up: your Ghidra DB still has these globals under the `rdMatrix44_unk*` / `DAT_` names, so it will disagree with the branch until `importDataSymbols.py` runs.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -340,6 +340,7 @@ swrTextFmtString2 0x004c3e48 char[8] = "~f%d%s\0\0"
 DAT_004c4000 0x004c4000 int
 
 maybeUICameraPlacements 0x004C4010 swrUICameraPlacement[40]
+swrObjHang_wattoCounterEyePos 0x004C4558 rdVector3 # base eye position for the jittered Watto-counter framing
 
 drawDistance 0x004c4a5c float
 
@@ -708,7 +709,9 @@ swrObjHang_menuTextPulsePhase 0x0050c8f4 float # selected menu-item text-color p
 swrUI_localPlayersInputDownBitset 0x0050C908 int[4]
 swrUI_localPlayersInputPressedBitset 0x0050C918 int[4]
 
-DAT_0050c930 0x0050c930 int
+swrObjHang_cameraMoveMode 0x0050c930 short # front-end camera transition mode; every access is 16-bit, so 0x0050c932 stays free
+swrObjHang_cameraMoveParity 0x0050c934 int # flips 0<->1 each time a camera move completes
+swrObjHang_cameraMoveActive 0x0050c940 int # a camera move is in flight
 
 swrObjHang_fadeState 0x0050c944 int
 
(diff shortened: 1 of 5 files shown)


louriccia:
Follow-up stacked on this: #294 reimplements the three functions that actually drive these matrix triples (the duration-based lerp, the speed-based step, and the shared step helper). Worth merging this one first — #294's diff includes this commit.

PR #294, opened 2026-09-02
reimpl(swrObjHang): the front-end camera move engine

louriccia:
Stacked on #293 - merge that first; the first commit is its. Review only `942947c` (the second commit).

Reimplements the three functions that actually drive the eye / look-at matrix triples #293 names.

| fn | addr |
|---|---|
| `swrObjHang_LerpHoloCamera` | `0x0045c0b0` |
| `swrObjHang_UpdateHoloCamera` | `0x0045c3c0` |
| `swrObjHang_StepCameraToward` | `0x0045c560` |

### Two move implementations, one completion path

The interesting bit is that the front end has **two independent camera moves**:

- **`LerpHoloCamera` is duration-based.** It snapshots the six move deltas, runs a 0.5s clock -- 0.3s for a short hop in the junkyard -- and lerps both rows. Mode 3 moves the look-at only and leaves the eye put.
- **`UpdateHoloCamera` is speed-based.** It delegates each row to `StepCameraToward`, which carries a speed between frames and eases in over a sqrt decel curve, then applies the pending menu state once both rows arrive.

`StepCameraToward` has to be in the same PR -- `UpdateHoloCamera` calls it, so shipping it as a prototype would break the link.

### `swrObjHang_screenTransitionDir` `0x0050c468` is not a boolean

Worth flagging because it reads like a flag and is not one. It is a **signed step**:

- `SetMenuState` clears it in the same block where it commits a state
- `UpdateScreenTransition` sets it to `1` and then *multiplies by it* to drive the slide offset
- `UpdateJunkyard` sets it to `-1` to run the slide in reverse

Nonzero means a screen slide is in flight, which is why a finished camera move refuses to apply its pending menu state while it is set. I held off naming this one until all four users lined up.

### Notes

The six lerp deltas are **individual interleaved floats, not two `rdVector3`s** -- look-at lives at `8b4`/`8c4`/`8e4` and eye at `8b8`/`8b0`/`8bc`, so they cannot be declared as vectors.

`swrObjHang_cameraStepScratch` `0x0050c9a4` is a dead store: `StepCameraToward` zeroes it and nothing in the binary reads it. Kept for faithfulness, named so it is obvious.

The `menuScreen` comparisons map onto the existing `swrObjHang_STATE_{PLANET_SELECT,RESULTS,VEHICLE_SELECT}_INTRO` members, so no new enum. Constants decode to 5.0 / 0.2 / 60000.0 / 0.05 / 6.0 / -2000.0.

Builds and links clean; hook-annotation, header-dup and data-symbol-dup checks pass. Burndown 895 -> 898.

### DIFF
--- a/data_symbols.syms
+++ b/data_symbols.syms
@@ -340,6 +340,7 @@ swrTextFmtString2 0x004c3e48 char[8] = "~f%d%s\0\0"
 DAT_004c4000 0x004c4000 int
 
 maybeUICameraPlacements 0x004C4010 swrUICameraPlacement[40]
+swrObjHang_wattoCounterEyePos 0x004C4558 rdVector3 # base eye position for the jittered Watto-counter framing
 
 drawDistance 0x004c4a5c float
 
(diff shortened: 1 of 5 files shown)


louriccia:
One more on top: #295 adds the retarget + idle sway (the last two small ones in this family). Merge order is #293 -> #294 -> #295.

PR #299, opened 2026-09-04
reimpl(swrRace): profile working <-> saved slot copies

louriccia:
Stacked on #298 — merge that first; the first commit here is its. Own range is 2 files, 17 lines.

`swrRace_CopyProfileFromSave` / `CopyProfileToSave` move one 0x50-byte profile record between the working array at 0x00e35a60 and the saved array at 0x00e364b4. Both were inlined 0x14-dword copies, so they become `memcpy` the way `swrRace_ApplyUpgradesToStats` already does.

The saved block gets its name here, and it cross-validates #298 from the other direction: the seven `traction/turning/…_upgrade_level` bytes already named at 0x00e364f5 sit at +0x41 from 0x00e364b4 — exactly the offset the garage preview reads out of the *working* profile. Two arrays, one 0x50-byte record layout, arrived at independently. `swrRace_SaveProfile` writes a single such record (0x10003 header + 0x50 bytes) and takes its pointer as the player name, which puts the name at the start of the record.

Also names the per-profile bytes at +0x21 / +0x22 — a flag and a saved-slot index that `swrRace_SaveCurrentProfile` and `swrObjHang_UpdateEnterName` use to mirror a working profile into a second saved slot. Both index with the same stride 0x50 as the existing `swrRace_UnlockDataBase` byte.

`swrRace_SaveCurrentProfile` stays stubbed on purpose: it calls `swrRace_SaveGameData` and `swrRace_SaveProfile`, neither of which has a body in `src/` yet, so reimplementing it now wouldn't link.

Full dinput.dll builds and links clean.


louriccia:
Superseded by #273, which reimplemented both `swrRace_CopyProfileFromSave` (0x0044e500) and `swrRace_CopyProfileToSave` (0x0044e530) at the same addresses — and better: it types `swrRace_aProfiles` as `swrSaveProfile[20]` and does a plain struct assignment, where this PR was still doing `memcpy` with hand-rolled offset constants.

Nothing left in the diff, so closing. The stack parent #298 is unaffected and is resynced against current master.

PR #308, opened 2026-09-10
Loose per-model replacement (RAWM chunk loader)

louriccia:
Lets a single model be replaced from a loose file, instead of shipping a whole rebuilt `out_modelblock.bin`.

Drop a chunk at `./assets/replacement_models/<model_id>.bin` and that model loads from it: the loader assembles a single-entry modelblock in memory, points the block path at it, remaps the id, and restores the path afterwards. Format is `"RAWM"` + mask/model sizes + the two payloads verbatim (big-endian, exactly the byte ranges that entry occupies in the stock block), so no conversion is involved in either direction.

Two scripts come with it: `extract_raw_model.py` carves a chunk out of a block (so the round trip is checkable -- extract model N, drop it back as N, nothing should change), and `validate_raw_model.py` checks the container, walks the mask-flagged words, and reports pointer/texture-index problems before the game ever sees the file.

The renderer change normalizes the blender-swe1r vertex-load encoding (`n=0, v0_plus_n=count`) to the game's (`n=count, v0=base`) where the display list is actually read. Without it a blender-authored chunk hits the `v0 != vertex_base_offset` abort; it is a no-op for game-native data, which never uses that encoding.

This is groundwork for shipping custom tracks as discrete assets rather than repacked archives -- a pack that redistributes the whole block is both large and forces the loader to guess which entry is the custom one by hashing.

Rebased onto current master (it had gone stale) and playtested two ways on the planet-select hologram (`MODELID_pln_tatooine_part`): a blender-authored chunk, which exercises the vertex-encoding path -- 42 of its vertex loads use `n=0` where a stock carve has none -- and a byte-exact stock carve, which renders indistinguishably from vanilla. A control run with the replacement removed logs nothing and loads from the block as before. Built + linked clean (WinLibs GCC 13.2).

### DIFF
--- a/dinput_hook/game_deltas/swrModel_delta.cpp
+++ b/dinput_hook/game_deltas/swrModel_delta.cpp
@@ -12,6 +12,7 @@
 #include "../node_utils.h"
 #include "../stb_image.h"
 #include "../custom_tracks.h"
+#include "../model_replacement.h"
 #include "../nv_dds/nv_dds.h"
 #include "../imgui_utils.h"
 #include "../ui_transform.h"
@@ -214,6 +215,7 @@ void swrText_InitFonts_delta(void) {
 swrModel_Header *swrModel_LoadFromId_delta(MODELID id) {
     const MODELID requested_id = id;
     const bool is_custom_track = prepare_loading_custom_track_model(&id);
+    const bool is_loose_model = !is_custom_track && try_prepare_loose_model(&id);
 
     char *model_asset_pointer_begin = swrAssetBuffer_GetBuffer();
     swrModel_Header *header = hook_call_original(swrModel_LoadFromId, id);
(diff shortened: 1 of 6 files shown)


louriccia:
Bumping this one for merge order rather than for anything new in it — it's unchanged, rebased on master and green.

#308 is the bottom of a nine-deep chain now: #309 → #310 → #311 → #312 → #313 → #314 → #315 → #316 are each cut from the one below, so their diffs are cumulative and none of them can land until this does. Reviewing it in isolation is also the smallest ask of the nine — it's just the RAWM chunk loader for loose per-model replacement, no catalog, no network, no UI.

If the loader shape looks right to you, merging this unblocks the whole line and I'll rebase the rest down as it goes.

PR #317, opened 2026-09-15
Music, ambient cues and cinematic from the manifest (stock slot, step 2)

louriccia:
Stacked on #316 - merge that first; this PR's own range is the last two commits.

**Draft: step 2 of retiring `overrides_stock_slot` - the four plain-table consumers.**

The EXE looks four things up in `.data` tables indexed by `(planet, subtrack)`: the in-race music (`swrMusicTrackTable @0x4b8750`), the planet's preload theme (`swrMusicPlanetIntroTable @0x4b8780`), the ambient cue list (`swrSfxPreloadSets @0x4b8fa8`) and the pre-race cinematic (`@0x4b7a08`, now named `swrPlanetIntroCinematics` in `data_symbols.syms`). This makes all four settable per track from the manifest.

None of the readers are touched. The registry writes the track's values into the entries its planet and subtrack index while the track is current, and restores them on the next track change, so a stock track always sees the originals. That is the whole of "shape 1" in the plan - the cheap shape - and it needed no new RE: the vanilla tables were read straight off the EXE's `.data` and match the reversed notes byte for byte.

```json
"environment": {
  "inherit": "vanilla:track:0",
  "music": "sebulbaloop", "intro_music": "mx091lavacaves", "cutscene": "none",
  "ambient": [{ "sound": "sfx_amb_crowd_big_loop", "start": 0.0, "end": 1.0, "mode": "loop" }]
}
```

Sounds are named, not numbered: `sound_map` parses `data/Sounds.map` the way the game does (0-based over active entries - `podloop1` = 144 and `DroidLoop1` = 152 land where the EXE's table expects them). `"cutscene": "none"` skips the planet intro for this track only, checked in `Window_PlayCinematic_delta`; the startup movies are not the track's to skip.

Two deliberate limits: a sound must already be in the game's bank (custom `.wav`s through the content store need a bank-registration path that does not exist yet - next slice), and `subtrack == 3` tracks are special-cased inside the readers rather than table-driven, so they cannot be overridden this way.

Playtested: Mute City (inherits Boonta Training's slot) with all four overrides - Sebulba's loop in race, the lava-caves intro theme, no fly-in, crowd all lap plus a random wind cue - then Boonta Training itself back to vanilla music, intro and movie.


louriccia:
Folded into #316 - one descriptor story, one review.

PR #318, opened 2026-09-15
Fog, draw distance and AI difficulty from the manifest (stock slot, step 3)

louriccia:
Stacked on #317 - merge that first; this PR's own range is the last commit.

**Draft: step 3 of retiring `overrides_stock_slot` - the two consumers that keep their data inside a function.**

**AI.** `InitAISettingsForTrack @0x4667e0` builds its (level, spread) table on the stack and picks the scripted-AI id and spline variant by planet and subtrack. The delta keeps that table and those rules exactly, then lets the descriptor override any of the four: `"ai": {"level": 12.0, "spread": 8, "script": -1, "spline_variant": 0}`. Beyond authoring difficulty, this fixes something real: a custom track borrowing Ando Prime's slot currently inherits Ando's *scripted AI behaviour* (control flow keyed on `planetId == 1`); `"script": -1` turns it off. The in-race restart path in the delta layer called the bare symbol, which under `hook_replace` would have run the dormant reimpl - it now calls the delta.

**Fog and draw distance.** `swrObjJdge_SetupTrackEnvironment @0x464b90` decides the fly-by spline, draw distance, fog start and fog/clear colour in a nested switch. It is not reimplemented: the original runs and the descriptor has the last word on the draw distance and the fog, which are plain outputs (`"draw_distance": 3000, "fog": {"near": 985, "color": [70, 20, 110]}` or `"fog": false`). That avoids naming the four unnamed globals a full reimpl would need first (`DAT_00e288b4`, `DAT_004c4a60`, `DAT_004ad22c`, `DAT_00e996e0` - noted for later). The fly-by spline is loaded inside the function so it is not overridden here, and does not need to be: a track that wants its own declares a spline asset at the inherited slot's camera-spline index, which the block view already serves.

A `HANG` stub for `SetupTrackEnvironment` in `swrObj.c` gives the hook an address to take (header-only callee). `JDGE_FLAG_SPECIAL_EVENT (0x20)` mirrors the raw `0x20` the existing reimpl tests; it wants a `types_enums.h` home once the judge flag set is named.

Playtested: Mute City purple-fogged with a 3000 draw distance and a harder, tighter AI field; Boonta Training afterwards back to sandy fog and normal AI; pause-menu Exit clean with both hooks live.


louriccia:
Folded into #316 - one descriptor story, one review.

